PR #40, opened 2013-04-21
[JENKINS-14155] Make it possible to automatically select/build the latest svn tag

pmoranga:
With this change we are able to build the last tag matching the filters set. For that you need to:
- Set defaultTagParameter to match the desired tag.
- and maxTags is 1.


jglick:
Merge conflicts. Anyway this plugin needs a lot of cleanup before new features will be considered, unless they are clearly self-contained.

kuisathaverat:
Is https://github.com/jenkinsci/subversion-plugin/pull/157 the same use case?

PR #44, opened 2013-07-02
introduce abstract class for post commit hook to decouple the payload processing from hook trigger logic

ndeloof:
allows to implement alternate web-hook payload computation
introduced for cloudforge-plugin development, see https://github.com/jenkinsci/cloudforge-plugin/issues/3
could be used as well to support googlecode post-commit hook payload


jglick:
Obsolete?

oleg-nenashev:
Yes, it would be hard to move it forward.
@ndeloof WDYT?

PR #64, opened 2014-02-01
JENKINS-5347 Fixed - Added use commit times on files

redsolo:
This fixes JENKINS-5347, we have a third party tool that uses the
timestamp on files to notify a cache in the tool to be updated. This fix
will add an option to the Subversion configuration to use commit times.
As the "use-commit-times" is a global configuration but we are only
interested in it as a per-project configuration to not affect our other
builds.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,7 +143,13 @@ THE SOFTWARE.
       <version>1.5</version>
       <scope>test</scope>
     </dependency>
-
+    <dependency>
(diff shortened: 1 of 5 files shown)


recena:
@redsolo Hello. Do you agree if we close this PR? It seems that there is a PR more recent and based on this. This PR is https://github.com/jenkinsci/subversion-plugin/pull/116

PR #72, opened 2014-03-28
[FIXED JENKINS-6746] Updated to add setting of SVN_MAX_REVISION env

srice01:
As per JENKINS-6746, the revision returned in SVN_REVISION is the revision from trunk but not the "current revision". This change sets a new "SVN_MAX_REVISION" environmental variable to be the value of the "current revision" (a maximum of any revisions involved in the repository including any externals). This allows, for example, the setting of the build number to be the current revision so that it is unique if externals, for example, are updated but not trunk.

One thing to note is that if you have multiple SVN locations (different checkouts in the project) then as there is no way to differentiate in "revisions.txt" between these locations so this variable is set to the maximum revision present across all checked-out SVN repositories.

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -716,6 +716,11 @@ public class SubversionSCM extends SCM implements Serializable {
                 }
             }
 
+            // get the maximum revision which attempts to resolve JENKINS-6746
+            Long maxrev = new Long(0);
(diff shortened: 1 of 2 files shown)


t-stamm:
Any news on this one?

recena:
@t-stamm I hope to back here soon.

jglick:
Same as #105?

Eccenux:
Implemented in: Added SVN_REVISION_MAX to build environment variables - #1598

A bit different naming, but same function.

PR #75, opened 2014-04-21
Avoid NullPointerException on FishEye CR3.1.5 FE3.1.5 (20131030082910 20...

Ntropy:
If we using url http://jira/browse/<repo>/<file>?r1=50647&r2=50648 we get NullPointerException on FishEye CR3.1.5 FE3.1.5 (20131030082910 2013-10-30)
For http://jira/browse/<repo>/<file>?r2=50648 we get expected result, no error.


oleg-nenashev:
@Ntropy are you interested in it? I doubt, but please respond to the comments above if yes

PR #83, opened 2014-05-13
[FIX JENKINS-23007] Consider environment when polling.

daniel-beck:
Otherwise, the credentials for locations with variables in their URL
aren't considered, requiring a fallback to Additional Credentials.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.489</version>
+    <version>1.509</version>
   </parent>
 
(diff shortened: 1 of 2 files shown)


recena:
@daniel-beck I've thought it would be interesting to close PR. I've filled a [Jira ticket](https://issues.jenkins-ci.org/browse/JENKINS-29340) and I've sent a new [PR](https://github.com/jenkinsci/subversion-plugin/pull/127) in order to resolve the same problem. Do you agree?

recena:
@daniel-beck This PR isn't valid because [this method](https://github.com/daniel-beck/subversion-plugin/blob/JENKINS-23007/src/main/java/hudson/scm/SubversionSCM.java#L1336) was refactoring recently.

PR #84, opened 2014-05-24
Show links to model object in polling log

daniel-beck:
Use the full name of the project. Link the node and (somewhat redundantly) the project from the polling log.

### DIFF
--- a/src/main/java/hudson/scm/CompareAgainstBaselineCallable.java
+++ b/src/main/java/hudson/scm/CompareAgainstBaselineCallable.java
@@ -1,6 +1,8 @@
 package hudson.scm;
 
+import hudson.console.ModelHyperlinkNote;
 import hudson.model.Hudson;
+import hudson.model.Node;
 import hudson.model.TaskListener;
 import hudson.remoting.DelegatingCallable;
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/scm/CompareAgainstBaselineCallable.java
```diff
@@ -54,7 +57,9 @@ public ClassLoader getClassLoader() {
      * so
      */
     public PollingResult call() throws IOException {
-        listener.getLogger().println("Received SCM poll call on " + nodeName + " for " + projectName + " on " + DateFormat.getDateTimeInstance().format(new Date()) );
+        String nodeLink = ModelHyperlinkNote.encodeTo("".equals(nodeName) ? ((Node) Jenkins.getInstance()) : Jenkins.getInstance().getNode(nodeName));
```
jglick: This can be run on the agent JVM, in which case this code will throw NPEs or worse. You would need to do the note encoding on the master side and pass the result as `nodeName`. Ditto for `projectName`.

daniel-beck:
I am no longer interested in completing this.

PR #88, opened 2014-06-19
[FIXED JENKINS-23498] - A new option, which disables the changelog calculation

oleg-nenashev:
Sometimes the SVN checkout and changelog generation procedures take much time due to the big number of <code>svn:external</code> dependencies.

Our use-case:
- checkout = 10min
- changelog calculation = 15min

The proposed option can be used to temporarily disable the calculation of changelogs in such case (e.g. for the job development purposes).

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -258,6 +258,10 @@ public class SubversionSCM extends SCM implements Serializable {
     private Boolean doRevert;
 
     private boolean ignoreDirPropChanges;
+    /**
+     * @since TODO: define a version
+     */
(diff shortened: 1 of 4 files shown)


jglick:
And it is already an option in Pipeline.

oleg-nenashev:
@reviewbybees just to decide something about that

jglick:
-0 as before

recampbell:
Possible improvement is to add a TODO saying this work is temporary pending changes in core.

jglick:
A user configuration option once added is hard to remove.

oleg-nenashev:
There is no community votes regarding it && I have no interest in delivering it. Hence just closing. The PR will be still linked in the ticket, so somebody can recover it if he wants

PR #89, opened 2014-07-02
[FIXED JENKINS-22568] Subversion polling does not work with parameters.

christ66:
Currently subversion polling does not work with parameters. This code changes polls using the _DEFAULT_ option. If there is a choice parameter, it will choose the _FIRST_ option. This will also expand anything that is `$VAL` or `${VAL}`.


smruti-sahoo:
Hi - I'm facing the same issue with git poll SCM 
Global environment variable are not being picked up during poll SCM 

Could someone please help me to solve this issue ?

recena:
@smruti-sahoo Please, fill a Jira issue with a step by step to reproduce the bug.

smruti-sahoo:
@recena Thanks - I've raised [JENKINS-34603](https://issues.jenkins-ci.org/browse/JENKINS-34603)
I found a similar issue raised around a year before [JENKINS-26315] (https://issues.jenkins-ci.org/browse/JENKINS-26315) but no update there, hence raised a new one

PR #94, opened 2014-08-31
Remove modified svnkit library from subversion plugin.

christ66:
Currently the subversion plugin is using a modified version of the svnkit library. This PR switches to the default svnkit in order to make svnkit upgrading easier.


src/test/java/hudson/scm/AbstractSubversionTest.java
```diff
@@ -61,6 +62,23 @@ protected Proc runSvnServe(File repo) throws Exception {
...
+          // Since the default svnserve port is currently in use, fail the build.
+          fail("Port 3690 is currently in use. Please kill anything running on this port and re-run.");
+        } catch (IOException e) {
+          // Port is not in use
+        } finally {
+          if (s != null) {
+            s.close();
+          }
+        }
+
```
jglick: As noted in #161, this is wrong and causes test failures when you run with sufficient concurrency. Picking a different port would be fine if you were _using_ that port in the URL defined in the test case. But you are not. We just have for example `"svn://localhost/bob"` which assumes the default port of 3690.

PR #105, opened 2014-12-17
Exporting new SVN_REVISION_HIGHEST for projects with multiple svn modules

caspercba:
Objective: Be able to know whats the highest svn revision among all the svn modules contained in the project. The var name is SVN_REVISION_HIGHEST
Motivation: We need to name the final release files using the svn revision. The problem arises when the build was triggered by a change in one of the libraries. We had to find a way to list all the svn revisions for our project and choose the highest one. 
Cons: If a project is located in different svn repositories, then there's no point in comparing their revisions. This wouldn't be an issue but the SVN_REVISION_HIGHEST wouldn't be used.

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -677,7 +677,7 @@ public class SubversionSCM extends SCM implements Serializable {
         try {
             Map<String,Long> revisions = parseSvnRevisionFile(build);
             Set<String> knownURLs = revisions.keySet();
-	    Long revHighest = 0;
+	    Long revHighest = new Long(0);
             if(svnLocations.length==1) {
                 // for backwards compatibility if there's only a single modulelocation, we also set
                 // SVN_REVISION and SVN_URL without '_n'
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -677,6 +677,7 @@ public void buildEnvVars(AbstractBuild<?, ?> build, Map<String, String> env) {
         try {
             Map<String,Long> revisions = parseSvnRevisionFile(build);
             Set<String> knownURLs = revisions.keySet();
+	    Long revHighest = new Long(0);
```
oleg-nenashev: Use <code>long</code> instead of <code>Long</code>

oleg-nenashev:
The improvement looks to be reasonable for me

recena:
@caspercba Hello. Could you provide a use case? When you have configured several modules (repos) you can get `SVN_REVISION` environment variable for each one.

recena:
@caspercba Are you still interested in this PR?

caspercba:
@recena Yes we can get the svn revision for each one, but we also have to loop through those and find the highest. Looping through svn variables with unknown length its not ideal. So it would be great if jenkins could provide the highest value in one variable. 
I don't need this feature anymore since we don't use our own libraries in our current projects.

recena:
@caspercba I believe that this functionality is not responsibility of this plugin. if you agree, I'll close this PR. ok?

jglick:
Duplicate of #72?

caspercba:
@recena Please close it. We have no need for this, and it doesnt seem to be very popular.

PR #106, opened 2015-01-18
[FIXED JENKINS-26264] UpdateUpdater perform will throw an AbortException

tastatur:
The UpdateUpdater.perform will throw an AbortException now, as suggested in [this issue description](https://issues.jenkins-ci.org/browse/JENKINS-26264)

### DIFF
--- a/src/main/java/hudson/scm/subversion/UpdateUpdater.java
+++ b/src/main/java/hudson/scm/subversion/UpdateUpdater.java
@@ -26,6 +26,7 @@
  */
 package hudson.scm.subversion;
 
+import hudson.AbortException;
 import hudson.Extension;
 import hudson.model.Hudson;
 import hudson.scm.SubversionSCM.External;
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/scm/subversion/UpdateUpdater.java
```diff
@@ -167,7 +168,7 @@ private SVNInfo parseSvnInfo(File workspace) throws SVNException {
             } catch (SVNCancelException e) {
                 if (isAuthenticationFailedError(e)) {
                     e.printStackTrace(listener.error("Failed to check out " + location.remote));
-                    return null;
+                    throw (AbortException) new AbortException().initCause(e);
```
jglick: `AbortException` intentionally does not have a constructor initializing `cause` becausing calling `initCause` on it is kind of pointless: the whole reason for using `AbortException` is to not print a stack trace.

And printing the stack trace of the `SVNCancelException` is probably too noisy, when this is a normal enough mistake.

So the whole `if` block body should probably just read

    throw new AbortException("Failed to check out " + location.remote);

PR #107, opened 2015-01-30
[JENKINS-26264] Cleanup AbortException call.

christ66:
Cleanup `AbortException` call by removing unnecessary part. See comment https://github.com/jenkinsci/subversion-plugin/pull/106#discussion_r23236498.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/scm/subversion/UpdateUpdater.java
+++ b/src/main/java/hudson/scm/subversion/UpdateUpdater.java
@@ -166,8 +166,8 @@ public class UpdateUpdater extends WorkspaceUpdater {
                         break;
                 }
             } catch (SVNCancelException e) {
+                e.printStackTrace(listener.getLogger());
                 if (isAuthenticationFailedError(e)) {
-                    e.printStackTrace(listener.getLogger());
                     throw new AbortException("Failed to check out " + location.remote);
                 } else {
                     listener.error("Subversion update has been canceled");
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/scm/subversion/UpdateUpdater.java
```diff
@@ -167,8 +167,7 @@ private SVNInfo parseSvnInfo(File workspace) throws SVNException {
                 }
             } catch (SVNCancelException e) {
                 if (isAuthenticationFailedError(e)) {
-                    e.printStackTrace(listener.error("Failed to check out " + location.remote));
-                    throw (AbortException) new AbortException().initCause(e);
+                    throw new AbortException("Failed to check out " + location.remote);
```
oleg-nenashev: I would also add the cause to exceptions in both <code>if</code> and <code>else</code> clauses

daniel-beck:
Wouldn't it be better to fix handling of AbortException to match its documentation?

> Signals a failure where the error was anticipated and diagnosed. When this exception is caught, the stack trace will not be printed, and the build will be marked as a failure.

Then the separate printing of stack trace wouldn't be redundant.

christ66:
@daniel-beck Good point. I reverted based on your input, and removed the duplicate exception message.

oleg-nenashev:
@christ66 Do you plan to handle my comment or to respond to it?

recena:
@christ66 Do you plan to review this PR?

christ66:
@oleg-nenashev @recena fixed.

recena:
@christ66 Thanks so much.

PR #109, opened 2015-01-30
[FIXED JENKINS-22112] Show commiter name on build status page

petermarcoen:
As suggested in https://issues.jenkins-ci.org/browse/JENKINS-22112
Show commiter name on build status page

@daniel-beck was wondering whether this should be shown stating that it's supposed to be a short + succinct summary of changes

I will leave the final decision to the community.
Here is the before and after:

<b>Before:</b> http://imgur.com/mEYuGgq,taNVqUL
<b>After:</b> http://imgur.com/mEYuGgq,taNVqUL#1

### DIFF
--- a/src/main/resources/hudson/scm/SubversionChangeLogSet/digest.jelly
+++ b/src/main/resources/hudson/scm/SubversionChangeLogSet/digest.jelly
@@ -60,7 +60,7 @@ THE SOFTWARE.
         <j:forEach var="cs" items="${it.logs}" varStatus="loop">
           <li>
             ${cs.msgAnnotated}
-            (<a href="${rootURL}/${cs.author.url}/">${cs.author}</a>/<a href="${changesBaseUrl}changes#detail${loop.index}">${%detail}</a>
+            (<a href="${changesBaseUrl}changes#detail${loop.index}">${%detail}</a>
 
             <j:set var="cslink" value="${browser.getChangeSetLink(cs)}"/>
             <j:if test="${cslink!=null}">
@@ -68,6 +68,8 @@ THE SOFTWARE.
               <a href="${cslink}">${browser.descriptor.displayName}</a>
             </j:if>
             <j:text>)</j:text>
+	    <br/>
+	    by <a href="${rootURL}/${cs.author.url}/">${cs.author}</a>
           </li>
         </j:forEach>
       </ol>
--- a/src/main/resources/hudson/scm/SubversionChangeLogSet/digest.jelly
+++ b/src/main/resources/hudson/scm/SubversionChangeLogSet/digest.jelly
@@ -60,7 +60,7 @@ THE SOFTWARE.
         <j:forEach var="cs" items="${it.logs}" varStatus="loop">
           <li>
             ${cs.msgAnnotated}
-            (<a href="${changesBaseUrl}changes#detail${loop.index}">${%detail}</a>
+            (<a href="${rootURL}/${cs.author.url}/">${cs.author}</a>/<a href="${changesBaseUrl}changes#detail${loop.index}">${%detail}</a>
 
             <j:set var="cslink" value="${browser.getChangeSetLink(cs)}"/>
             <j:if test="${cslink!=null}">


daniel-beck:
Is this the raw user name, or the user's display name? It should be the latter, and that can be pretty long.

As an example, I've seen people with long and/or multiple and given and surnames, and organizations add the name of the department and type of employment to the LDAP user name, e.g. `INTERN Eduardo Antonio Gómez Iglesias (AB-CD/EFG-HIJ12-KL)`. Maybe this should even be placed on a second line. With multi-line commit messages, this wouldn't even blow up the summary that much. With a third link in parentheses, they seem crowded.

petermarcoen:
This is the full name as it also used on the "Changes" page.

I understand your concern about long names. I updated it with the committer's name on a seperate line.

Here is a screenshot with a multi-line commit message as well: http://imgur.com/CCMFzIP

petermarcoen:
Any update on this @daniel-beck or others?

daniel-beck:
How does this look for users that don't have a Jenkins account? It's the raw user name from SVN, right?

petermarcoen:
Users that are not logged in see the name exactly like logged in users do:

image

The name you see is the Full Name of the user as set in the "People"-configuration when you press "Configure":

image

If this name is not defined, the username is taken from svn.

The variable I am using is the exact same one as on the "Changes"-page:

image

This is the code in the "Changes"-page:

```
    <j:forEach var="cs" items="${it.logs}" varStatus="loop">
      <tr class="pane">
        <td colspan="3" class="changeset">
          <a name="detail${loop.index}"></a>
          <div class="changeset-message">
            <b>
              ${%Revision}
              <a href="${browser.getChangeSetLink(cs)}">${cs.revision}</a>
              by <a href="${rootURL}/${cs.author.url}/">${cs.author}</a>:
            </b><br/>
            ${cs.msgAnnotated}
          </div>
        </td>
      </tr>
```

And this is the code I added:

```
        <j:forEach var="cs" items="${it.logs}" varStatus="loop">
          <li>
            ${cs.msgAnnotated}
            (<a href="${changesBaseUrl}changes#detail${loop.index}">${%detail}</a>

            <j:set var="cslink" value="${browser.getChangeSetLink(cs)}"/>
            <j:if test="${cslink!=null}">
              <j:text>/</j:text>
              <a href="${cslink}">${browser.descriptor.displayName}</a>
            </j:if>
            <j:text>)</j:text>
	    <br/>
	    by <a href="${rootURL}/${cs.author.url}/">${cs.author}</a>
          </li>
        </j:forEach>
```

I hope this answers your questions.

daniel-beck:
I meant users that exist in SVN but not in Jenkins (as users). This should default to the raw user name then for the virtual (SCM) user accounts AFAICT.

Anyway, this looks like a useful change. :+1:

petermarcoen:
@daniel-beck, I think this is the default behavior. If I look in the People tab I read the following:

image

I also noticed that users that commited something but that were not registered in Jenkins were added here with their SCM user name.

petermarcoen:
@daniel-beck and others: Any other remarks on this or can it be merged?

oleg-nenashev:
In my opinion, you should check if the user exists in Jenkins database before providing a direct link. Old Jenkins versions create users on-demand when you use the link.

daniel-beck:
@oleg-nenashev Subversion plugin has a fairly new core requirement, and any compatible release no longer has that issue IIRC.

oleg-nenashev:
@daniel-beck 
The issue has been partially fixed. Admins are still able to create new users by following a direct link

daniel-beck:
@oleg-nenashev Right, admins only, and those user records exist only until restart.

However, any user name linked here already has the user record created from the SCM changelog, right?

oleg-nenashev:
> Right, admins only, and those user records exist only until restart.

If an admin opens and  saves the config, they exist infinitely

> However, any user name linked here already has the user record created from the SCM changelog, right?

Well, my initial comment was not exact enough.

Changelog entries must return nonull <code>User</code>, but nothing blocks them from returning <code>User.getUnknown()</code> if they filter users somehow. Unknown user returns a non-persistent config page, which allows to keep properties till the restart. I would not like to expose it due to potential securtity breaches.

petermarcoen:
@oleg-nenashev and @daniel-beck Note that this is the same exact link as used on the Changes-page so if this is indeed a potential security breach, we should fix in in both places.

daniel-beck:
There is no security issue here. @oleg-nenashev This is how the API has worked forever. See any Jenkins' *People* page (preferably one building non-internal projects). As soon as they are author of a commit, they are in Jenkins, no clicking on them required (and only admins can set their `UserProperty`s if they deliberate choose to). Changing this behavior is way out of scope of this change.

oleg-nenashev:
@petermarcoen @daniel-beck 
I agree to keep the link as is

:+1:

petermarcoen:
So @daniel-beck, can we merge this?

recena:
@petermarcoen Nice work! Did you try your PR with latest changes in the `master` branch?

petermarcoen:
Thanks @recena, I tried it just now with the latest jenkins and the latest subversion-plugin, looks ok!

PR #111, opened 2015-02-15
added TortoiseSVN repository browser

ChakaZulu:
- TortoiseSVN must be installed on client machine
- client must be allowed to access repository
- client must allow execution of tsvncmd URL handler

### DIFF
--- a/src/main/java/hudson/scm/browsers/TortoiseSVN.java
+++ b/src/main/java/hudson/scm/browsers/TortoiseSVN.java
@@ -1,0 +1,194 @@
+/*
+ * The MIT License
+ * 
+ * Copyright (c) 2015 Michael Schuele.
+ * 
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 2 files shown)


daniel-beck:
It appears you're using the job configuration to determine where to link, rather than what has actually been checked out in a build. This leads to a number of possible issues (haven't tried it, but these are my guesses based on reading the code):

* This may not handle configuration changes well (i.e. when the URLs configured in the project don't match the URLs checked out in a build).
* This may not handle variables in configured SVN URLs (e.g. when using the Subversion parameter)
* This appears to not link to URLs included from externals
* This may not handle the `.` local checkout path (checking out directly into the workspace)

FWIW I've used https://github.com/daniel-beck/jenkins-tortoisesvn-browser-plugin successfully for more than a year and its behavior appears to be reasonably correct, so I could just release it if there's demand for it.

ChakaZulu:
Thanks for reviewing the changes.
1) yes, configuration changes are not covered, e.g. checking out into $workspace\subversion and changing that to $workspace\myRepo later would be a problem. I guess I just didn't find the correct information in SubversionChangeLogSet and took its preparePath method as an example. I thought the source repository path is just lost.
2) variables and externals: well, you are right, I didn't consider this :-)
3) that should work but anyway...

... I'll happily check out your plugin

ChakaZulu:
closing pull request because another plugin already exists

PR #112, opened 2015-03-12
[FIXED JENKINS-26668] Fixing 'Tag this build' credentials

rozboris:
Fix for JENKINS-26668 (https://issues.jenkins-ci.org/browse/JENKINS-26668)
For some reason, credential selection dropdown name is "_.credentialsId" in the multipart data. I think it might be encoded, so the real bug is in the MultipartFormDataParser, but it's a harmless workaround and it fixes the problem.

### DIFF
--- a/src/main/java/hudson/scm/SubversionTagAction.java
+++ b/src/main/java/hudson/scm/SubversionTagAction.java
@@ -230,6 +230,9 @@ public class SubversionTagAction extends AbstractScmTagAction implements Describ
         }
 
         String credentialsId = parser.get("credentialsId");
+        if(credentialsId == null){
+            credentialsId = parser.get("_.credentialsId");
+        }
         StandardCredentials upc = null;
         if (credentialsId != null) {
             Item context = req.findAncestorObject(Item.class);


rozboris:
Regarding the failed test (see https://jenkins.ci.cloudbees.com/job/plugins/job/subversion-plugin/487/) it can not have any connection to my change, something is fishy with the setup. Sorry, I don't have time to check it myself, but I'm sure my commit can not break anything.

rompic:
thank you for the fix. I will test this on our instance tomorrow

rompic:
I encountered https://issues.jenkins-ci.org/browse/JENKINS-26458 after the update from 2.0. But it works after setting the Client Version to 1.8 in Jenkins -> Config

Hawk777:
Thank you! I don’t have a whole build environment set up, so I didn’t actually use your patch, but it *did* help by pointing me in the right direction of which element to edit in the page so I could get my instance to tag. Hope to see this fixed properly by the next time I need to use it.

recena:
@rozboris This [Jelly file](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/lib/form/select.jelly#L52) can help you to understand the cause of the problem

christ66:
I am closing this pull request and will merge https://github.com/jenkinsci/subversion-plugin/pull/121 which contains a similar fix, and steps to reproduce this issue.

PR #113, opened 2015-03-17
WIP [JENKINS-5750] Get svn info from relative location instead of root repository

christ66:
WIP because I need a test unit. 

[https://issues.jenkins-ci.org/browse/JENKINS-5750](https://issues.jenkins-ci.org/browse/JENKINS-5750).

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -993,7 +993,7 @@ public class SubversionSCM extends SCM implements Serializable {
 
         private void checkClockOutOfSync() {
             try {
-                SVNDirEntry dir = clientManager.createRepository(location.getSVNURL(), true).info("/", -1);
+                SVNDirEntry dir = clientManager.createRepository(location.getSVNURL(), true).info("", -1);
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -993,7 +993,7 @@ public CheckOutTask(Run<?, ?> build, SubversionSCM parent, ModuleLocation locati
 
         private void checkClockOutOfSync() {
             try {
-                SVNDirEntry dir = clientManager.createRepository(location.getSVNURL(), true).info("/", -1);
+                SVNDirEntry dir = clientManager.createRepository(location.getSVNURL(), true).info("", -1);
```
oleg-nenashev: Comment in line 1004 will become invalid

oleg-nenashev:
weak :bug: . The change may break much other use-cases relying on SVN root.
I see two options:
* make the behavior configurable via advanced options
* use a relative path as a fallback if the Auth error happens for the root path

oleg-nenashev:
@christ66 do you have any plans to work on it?

PR #114, opened 2015-03-18
Added backward-compatibility for the svn-tag plugin

rozboris:
The svn-tag plugin (https://wiki.jenkins-ci.org/display/JENKINS/Subversion+Tagging+Plugin) was failing with MethodNotFound exception, so I had to implement this fix, adding backwards-compatible wrapper methods

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -463,6 +463,13 @@ public class SubversionSCM extends SCM implements Serializable {
     public ModuleLocation[] getLocations(AbstractBuild<?,?> build) {
         return getLocations(null, build);
     }
+    
+    /**
+     * Fix for the backward-compatibility with svn-tag plugin
+     */
+    public ModuleLocation[] getLocations(EnvVars env, AbstractBuild<?,?> build) {
+        return getLocations(env, (Run<?,?>)build);
+    }
 
     /**
      * List of all configured svn locations, expanded according to all env vars
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -1106,7 +1106,14 @@ public class SubversionSCM extends SCM implements Serializable {
         return CredentialsSVNAuthenticationProviderImpl.createAuthenticationProvider(inContextOf, this, location);
     }
 
-
+    @Deprecated
+    public ISVNAuthenticationProvider createAuthenticationProvider(AbstractProject<?,?> inContextOf, ModuleLocation location) {
+        SubversionSCM scm = null;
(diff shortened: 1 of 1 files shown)


rompic:
i have the same issue. I'll try to test this on friday

rompic:
works like a charm. thanks

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -463,6 +463,13 @@ public void setWorkspaceUpdater(WorkspaceUpdater workspaceUpdater) {
     public ModuleLocation[] getLocations(AbstractBuild<?,?> build) {
         return getLocations(null, build);
     }
+    
+    /**
+     * Fix for the backward-compatibility with svn-tag plugin
```
oleg-nenashev: "Fix for..." is a good remark, but not a summary
add <code>@since</code>. If the method existed before, it worth to mention it in remarks

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -1099,7 +1106,14 @@ public ISVNAuthenticationProvider createAuthenticationProvider(Job<?, ?> inConte
...
     }
 
-
+    @Deprecated
+    public ISVNAuthenticationProvider createAuthenticationProvider(AbstractProject<?,?> inContextOf, ModuleLocation location) {
+        SubversionSCM scm = null;
+        if (inContextOf != null && inContextOf.getScm() instanceof SubversionSCM) {
+            scm = (SubversionSCM)inContextOf.getScm();
+        }
+        return CredentialsSVNAuthenticationProviderImpl.createAuthenticationProvider(inContextOf, scm, location);
```
oleg-nenashev: Passing null SCM may be risky, because the behavior is not documented. BTW, the implementations handle this null, so it's not very important

oleg-nenashev:
:+1: with minor comments

recena:
I think we can close this PR because It was solved in [Subversion Tagging Plugin 1.18](https://wiki.jenkins-ci.org/display/JENKINS/Subversion+Tagging+Plugin#SubversionTaggingPlugin-History)

christ66:
The backwards compatibility changes still need to be made. Another plugin which has a dependency on this plugin might require these deleted methods.

src/main/java/hudson/scm/SubversionSCM.java
christ66: Please also add an `@deprecated` to this method.

```
@deprecated Use {@link #getLocations(EnvVars, Run)}
```

recena:
@christ66 Any way, "those" plugin will need to update its dependency. And 1.18 has solved the problem. IMO, this PR don't make sense.

PR #117, opened 2015-06-13
[FIXED JENKINS-27084] Revert system property change

christ66:
Revert a code change which made the System properties `svnkit.http.methods`, and `svnkit.ssh2.persistent ` not get properly set.

@reviewbybees


stephenc:
:+1: after you fix the build

daniel-beck:
:+1:

PR #118, opened 2015-06-14
removed dependancy on codehaus

renea:
As codehaus has shut down.
Replace with jenkins server for variable tests.
Remove test for http checkout


daniel-beck:
#117 includes many of these changes as well.

renea:
Agree, #117 looks better as it does not remove the http checkout test.

PR #119, opened 2015-06-23
AbstractScmTagAction.getBuild is deprecated in favor of getRun

jglick:
Not sure exactly when the `tagForm` view is being used, but if it is ever used from a Workflow build (or any other non-`AbstractBuild`), this patch would be needed in order to refer to the build.

@reviewbybees

### DIFF
--- a/src/main/resources/hudson/scm/SubversionTagAction/tagForm.jelly
+++ b/src/main/resources/hudson/scm/SubversionTagAction/tagForm.jelly
@@ -41,9 +41,9 @@ THE SOFTWARE.
   <j:set var="descriptor" value="${it.descriptor}" />
 
   <l:layout norefresh="true" xmlns:local="local">
-    <st:include it="${it.build}" page="sidepanel.jelly" />
+    <st:include it="${it.run}" page="sidepanel.jelly" />
     <l:main-panel>
-      <h1>${%Build} #${it.build.number}</h1>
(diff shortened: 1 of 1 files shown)


varmenise:
[just asking to understand] if we care about changing getBuild in favour of getRun in jelly files, shouldn't we do the same in the java implementation (https://github.com/jglick/subversion-plugin/blob/master/src/main/java/hudson/scm/SubversionTagAction.java has several getBuild calls)

oleg-nenashev:
@varmenise you're right, there should be a big migration in the code to new APIs.

BTW, :+1: for this particular change

stephenc:
@oleg-nenashev you have forgotten we are not about the thumbs any more

stephenc:
:bee:

varmenise:
:bee:

oleg-nenashev:
(bee)

oleg-nenashev:
@stephenc 
I had a long 1-day vacation, so probably I've forgotten something  :)

stephenc:
@oleg-nenashev its the : form not the ( form of bee BTW

jglick:
> SubversionTagAction.java has several getBuild calls

Yes it does, but would require other changes to work with non-`AbstractBuild`s. In this case I was just making a sweep of forgotten references to `.build` from Jelly files, which the Java compiler does not flag, as part of fixing a real bug upstream in core which affected the `p4` plugin.

PR #120, opened 2015-07-03
[JENKINS-29211] E200015: ISVNAuthentication provider did not provide credentials

recena:
https://issues.jenkins-ci.org/browse/JENKINS-29211

This pull request tries to improve the migration process that the plugin implements for upgrading repositories with an old credentials format.

Result after upgrading:

jenkins-29211_1

We lose the relation between our repository and the credentials.

### DIFF
--- a/src/main/java/hudson/scm/PerJobCredentialStore.java
+++ b/src/main/java/hudson/scm/PerJobCredentialStore.java
@@ -115,16 +115,13 @@ final class PerJobCredentialStore implements Saveable, RemotableSVNAuthenticatio
         CredentialsStore store = CredentialsProvider.lookupStores(project).iterator().next();
         for (Map.Entry<String, Credential> e : credentials.entrySet()) {
             StandardCredentials credential =  descriptor.migrateCredentials(store, e.getKey(), e.getValue());
-            int i = 0;
-            boolean found = false;
             ModuleLocation[] locations = ((SubversionSCM) project.getScm()).getLocations();
-            while (i < locations.length && !found) {
+            for (int i = 0; i < locations.length; i++) {
                 try {
                     if (e.getKey().contains(locations[i].getSVNURL().getHost())) {
                         locations[i].setCredentialsId(credential.getId());
-                        found = true;
+                        break;
                     }
-                    i++;
                 } catch (SVNException ex) {
                     // Should not happen, but...
                     LOGGER.log(WARNING, "Repository location with a malformed URL: " + locations[i].remote, ex);
--- a/src/main/java/hudson/scm/PerJobCredentialStore.java
+++ b/src/main/java/hudson/scm/PerJobCredentialStore.java
@@ -112,19 +112,22 @@ final class PerJobCredentialStore implements Saveable, RemotableSVNAuthenticatio
     private static final ThreadLocal<Boolean> IS_SAVING = new ThreadLocal<Boolean>();
 
     /*package*/ void migrateCredentials(SubversionSCM.DescriptorImpl descriptor) throws IOException {
-        CredentialsStore store = CredentialsProvider.lookupStores(project).iterator().next();
-        for (Map.Entry<String, Credential> e : credentials.entrySet()) {
-            StandardCredentials credential =  descriptor.migrateCredentials(store, e.getKey(), e.getValue());
-            ModuleLocation[] locations = ((SubversionSCM) project.getScm()).getLocations();
-            for (int i = 0; i < locations.length; i++) {
-                try {
-                    if (e.getKey().contains(locations[i].getSVNURL().getHost())) {
-                        locations[i].setCredentialsId(credential.getId());
-                        break;
+        Iterable<CredentialsStore> it = CredentialsProvider.lookupStores(project);
+        if (it != null && it.iterator().hasNext()) {
+            CredentialsStore store = it.iterator().next();
+            for (Map.Entry<String, Credential> e : credentials.entrySet()) {
+                StandardCredentials credential =  descriptor.migrateCredentials(store, e.getKey(), e.getValue());
+                ModuleLocation[] locations = ((SubversionSCM) project.getScm()).getLocations();
+                for (int i = 0; i < locations.length; i++) {
+                    try {
+                        if (e.getKey().contains(locations[i].getSVNURL().getHost())) {
+                            locations[i].setCredentialsId(credential.getId());
+                            break;
+                        }
+                    } catch (SVNException ex) {
+                        // Should not happen, but...
+                        LOGGER.log(WARNING, "Repository location with a malformed URL: " + locations[i].remote, ex);
                     }
-                } catch (SVNException ex) {
-                    // Should not happen, but...
-                    LOGGER.log(WARNING, "Repository location with a malformed URL: " + locations[i].remote, ex);
                 }
             }
         }
--- a/src/main/java/hudson/scm/PerJobCredentialStore.java
+++ b/src/main/java/hudson/scm/PerJobCredentialStore.java
@@ -4,10 +4,16 @@ import hudson.XmlFile;
 import hudson.model.AbstractProject;
 import hudson.model.Saveable;
 import hudson.remoting.Channel;
+import hudson.scm.SubversionSCM.ModuleLocation;
 import hudson.scm.SubversionSCM.DescriptorImpl.Credential;
 import hudson.scm.SubversionSCM.DescriptorImpl.RemotableSVNAuthenticationProvider;
-import jenkins.model.Jenkins;
+
+import org.tmatesoft.svn.core.SVNException;
 import org.tmatesoft.svn.core.SVNURL;
+
+import com.cloudbees.plugins.credentials.CredentialsProvider;
+import com.cloudbees.plugins.credentials.CredentialsStore;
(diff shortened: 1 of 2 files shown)


recena:
@reviewbybees Could you review this changes?

stephenc:
:bee:

src/main/java/hudson/scm/PerJobCredentialStore.java
```diff
@@ -105,8 +112,24 @@ private Object writeReplace() {
     private static final ThreadLocal<Boolean> IS_SAVING = new ThreadLocal<Boolean>();
 
     /*package*/ void migrateCredentials(SubversionSCM.DescriptorImpl descriptor) throws IOException {
+        CredentialsStore store = CredentialsProvider.lookupStores(project).iterator().next();
         for (Map.Entry<String, Credential> e : credentials.entrySet()) {
-            descriptor.migrateCredentials(project, e.getKey(), e.getValue());
+            StandardCredentials credential =  descriptor.migrateCredentials(store, e.getKey(), e.getValue());
```
tfennelly: So `store` can be passed as `null`? Or maybe `SubversionSCM.DescriptorImpl.migrateCredentials` can take `null` and react "properly"?

oleg-nenashev: Null handling in SVN plugin is quite random and indocumented. It would worth to spend several hours and to document core classes at least (out of the scope of this PR, of course)

src/main/java/hudson/scm/PerJobCredentialStore.java
```diff
@@ -105,8 +112,24 @@ private Object writeReplace() {
...
+            int i = 0;
+            boolean found = false;
+            ModuleLocation[] locations = ((SubversionSCM) project.getScm()).getLocations();
```
oleg-nenashev: Unchecked conversion (NIT in this case)

src/main/java/hudson/scm/PerJobCredentialStore.java
```diff
@@ -105,8 +112,24 @@ private Object writeReplace() {
...
+            while (i < locations.length && !found) {
```
oleg-nenashev: foreach cycle with break would be more readable

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -1792,8 +1792,7 @@ public SVNAuthentication createSVNAuthentication(String kind) {
 
             @Override
             public StandardCredentials toCredentials(String description) {
-                return new UsernamePasswordCredentialsImpl(CredentialsScope.GLOBAL, null, description, userName,
-                        getPassword());
+                return new UsernamePasswordCredentialsImpl(CredentialsScope.GLOBAL, null, description, userName, getPassword());
```
oleg-nenashev: please avoid refactorings if there's no particular need

src/main/java/hudson/scm/PerJobCredentialStore.java
```diff
@@ -105,8 +112,24 @@ private Object writeReplace() {
     private static final ThreadLocal<Boolean> IS_SAVING = new ThreadLocal<Boolean>();
 
     /*package*/ void migrateCredentials(SubversionSCM.DescriptorImpl descriptor) throws IOException {
+        CredentialsStore store = CredentialsProvider.lookupStores(project).iterator().next();
```
oleg-nenashev: My god-feeling that all credential stores should be checked

src/main/java/hudson/scm/PerJobCredentialStore.java
recena: @oleg-nenashev In this point, it is only possible ```SubversionSCM```.

oleg-nenashev: That's whyt it's NIT :)
A good practice would be to check the class and to throw IllegalState exception. Otherwise you'll get an issue from FindBugs

src/main/java/hudson/scm/SubversionSCM.java
recena: @oleg-nenashev It is not a refactoring, I removed only an unnecessary *carriage return*. In my opinion, is more legible.

oleg-nenashev: It's a refactoring. Any developer doing changes may suffer from merge conflicts

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -1663,9 +1664,8 @@ public void load() {
             save();
         }
 
-        /*package*/ StandardCredentials migrateCredentials(ModelObject context, String legacyRealm, Credential legacyCredential)
+        /*package*/ StandardCredentials migrateCredentials(CredentialsStore store, String legacyRealm, Credential legacyCredential)
```
oleg-nenashev: It may be a breaking change in the case of SVN plugin. There's a code within same package in other plugins AFAIK

src/main/java/hudson/scm/SubversionSCM.java
recena: @oleg-nenashev I considered it was necessary to do this.

src/main/java/hudson/scm/SubversionSCM.java
recena: https://github.com/search?q=migrateCredentials&type=Code&utf8=%E2%9C%93

src/main/java/hudson/scm/PerJobCredentialStore.java
recena: @oleg-nenashev https://issues.jenkins-ci.org/browse/JENKINS-29227

src/main/java/hudson/scm/PerJobCredentialStore.java
recena: @oleg-nenashev Done.

src/main/java/hudson/scm/SubversionSCM.java
oleg-nenashev: Such changes should be submitted as additional PRs, but actually it's unnecessary at all. The coding style does not explicitly define this requirement, and the text readability depends on your IDE window width.

src/main/java/hudson/scm/SubversionSCM.java
oleg-nenashev: Does it mean you disagree? I have no strong opinion, so it's up to you

src/main/java/hudson/scm/SubversionSCM.java
recena: @oleg-nenashev I'm not going to discuss if `line length` should o not be defined in a Coding Style Guide

src/main/java/hudson/scm/SubversionSCM.java
recena: @oleg-nenashev No I don't, it is only a reference.

christ66:
:bee: Testing this locally with 1.54 of the subversion plugin, and migrating to 2.50 I was able to reproduce the credentials not properly being migrated. Upgrading from 1.54 to this pull request I was able to maintain the relationship between previous credentials implementation to using the credentials plugin.

src/main/java/hudson/scm/PerJobCredentialStore.java
recena: @oleg-nenashev I'll do it.

src/main/java/hudson/scm/PerJobCredentialStore.java
```diff
@@ -105,8 +112,24 @@ private Object writeReplace() {
     private static final ThreadLocal<Boolean> IS_SAVING = new ThreadLocal<Boolean>();
 
     /*package*/ void migrateCredentials(SubversionSCM.DescriptorImpl descriptor) throws IOException {
-        for (Map.Entry<String, Credential> e : credentials.entrySet()) {
-            descriptor.migrateCredentials(project, e.getKey(), e.getValue());
+        Iterable<CredentialsStore> it = CredentialsProvider.lookupStores(project);
+        if (it != null && it.iterator().hasNext()) {
```
oleg-nenashev: foreach cycle would be better

recena: @oleg-nenashev But I only want the first.

amuniz:
:bee:

src/main/java/hudson/scm/PerJobCredentialStore.java
Vlatombe: guava's `Iterables#getFirst` may be an nice way to improve readability on this one.

recena: @Vlatombe Sure but this plugins does not use Guava library.

stephenc: @Vlatombe @recena 

FWIW I do not use Guava in Jenkins plugins unless there is a gun pointed at my head. It has caused more pain due to their differential view of backwards compatibility from that of the Jenkins project. I got badly burned in the 1.4xx versions of Jenkins due to Guava changes and had to rip out its usage. I recommend you avoid Guava in Jenkins plugins like one would avoid the plague

stephenc:
:bee:

recena:
@reviewbybees done

PR #124, opened 2015-07-13
Fix findbugs null pointer dereferences

christ66:
There are several locations where there could potentially be a null pointer exception.


src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
```diff
@@ -354,8 +355,16 @@ public ISVNAuthenticationProvider createAuthenticationProvider(AbstractProject c
...
     public FormValidation doCheckTagsDir(StaplerRequest req, @AncestorInPath Item context, @QueryParameter String value) {
-      return Jenkins.getInstance().getDescriptorByType(SubversionSCM.ModuleLocation.DescriptorImpl.class).doCheckRemote(req, context, value);
+        Jenkins instance = Jenkins.getInstance();
+        if (instance != null) {
+            SubversionSCM.ModuleLocation.DescriptorImpl desc = instance.getDescriptorByType(SubversionSCM.ModuleLocation.DescriptorImpl.class);
+            if (desc != null) {
+                return desc.doCheckRemote(req, context, value);
+            }
+        }
+        return null;
```
oleg-nenashev: Should return a warning IMO

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -2511,7 +2523,8 @@ private static boolean isAuthenticationFailedError(SVNCancelException e) {
     }
 
     private static DescriptorImpl descriptor() {
```
oleg-nenashev: It worth to add annotations to let FindBugs discover more issues later

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -2261,10 +2264,19 @@ protected void onSuccess(String realm, Credential cred) {
...
+            Jenkins instance = Jenkins.getInstance();
+            if (instance != null) {
+                ModuleLocation.DescriptorImpl d = instance.getDescriptorByType(ModuleLocation.DescriptorImpl.class);
+                if (d != null) {
+                    return d.doCheckCredentialsId(
+                            req, context, value, credentialsId);
+                }
+            }
+
+            return null;
```
oleg-nenashev: should be a warning

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -1644,19 +1644,22 @@ public void load() {
...
-            for (AbstractProject<?,?> job: Jenkins.getInstance().getAllItems(AbstractProject.class)) {
-                File jobCredentials = new File(job.getRootDir(), "subversion.credentials");
-                if (jobCredentials.isFile()) {
-                    try {
-                        new PerJobCredentialStore(job).migrateCredentials(this);
-                        job.save();
-                        if (!jobCredentials.delete()) {
-                            LOGGER.log(Level.WARNING, "Could not remove legacy per-job credentials store file: {0}", jobCredentials);
+            Jenkins instance = Jenkins.getInstance();
+            if (instance != null) {
```
oleg-nenashev: This approach causes a huge diff.
I would check Jenkins instance and pass empty array to the cycle if it is null

src/main/java/hudson/scm/SubversionRepositoryStatus.java
```diff
@@ -152,7 +155,8 @@ public void doNotifyCommit(StaplerRequest req, StaplerResponse rsp) throws Servl
         private JobProvider jobProvider = new JobProvider() {
             @SuppressWarnings("rawtypes")
             public List<Job> getAllJobs() {
-                return Jenkins.getInstance().getAllItems(Job.class);
+                Jenkins instance = Jenkins.getInstance();
+                return instance != null ? instance.getAllItems(Job.class) : new ArrayList<Job>();
```
oleg-nenashev: use <code>Collections.emptyList()</code>

src/main/java/hudson/scm/SubversionRepositoryStatus.java
```diff
@@ -132,13 +132,16 @@ public void doNotifyCommit(StaplerRequest req, StaplerResponse rsp) throws Servl
         }
 
         boolean listenerDidSomething = false;
-        for (Listener listener : Jenkins.getInstance().getExtensionList(Listener.class)) {
-            try {
-                if (listener.onNotify(uuid, rev, affectedPath)) {
-                    listenerDidSomething = true;
+        Jenkins instance = Jenkins.getInstance();
+        if (instance != null) {
```
oleg-nenashev: Use reverse approach to simplify diffs.

```java
Jenkins instance = Jenkins.getInstance();
if (instance == null) {
  return; // And log?
}
for (Listener listener : instance.getExtensionList(Listener.class)) {
...
```

oleg-nenashev:
It needs some polishing

oleg-nenashev:
:bee:

recena:
:bee:

recena:
@christ66 Do you think this PR is ready to be merged?

PR #125, opened 2015-07-13
Exclude unused subversion plugin dependency from jenkins-test-harness

christ66:
The jenkins-test-harness dependency had an initial exclusion of the subversion plugin, however since the plugins artifact id changed an exclusion needs to be made also.


oleg-nenashev:
:bee: , but probably such exclusions are not required anymore

christ66:
Writing unit tests I got confused as to which svnkit binaries I was using.

recena:
@christ66 That would be perfect!

jglick:
Merge conflict. I doubt the exclusion is still necessary at all.

PR #127, opened 2015-07-18
[JENKINS-29340] ISVNAuthentication provider did not provide credentials

recena:
https://issues.jenkins-ci.org/browse/JENKINS-29340

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -1392,7 +1392,13 @@ public class SubversionSCM extends SCM implements Serializable {
 
         final Map<String,ISVNAuthenticationProvider> authProviders = new LinkedHashMap<String,
                 ISVNAuthenticationProvider>();
-        for (ModuleLocation loc: getLocations()) {
+
+        EnvVars env = null;
+        if (project.getLastCompletedBuild() != null) {
+            env = project.getLastCompletedBuild().getEnvironment(listener);
+        }
+
+        for (ModuleLocation loc: getLocations(env, null)) {
             String url;
             try {
                 url = loc.getExpandedLocation(project).getSVNURL().toDecodedString();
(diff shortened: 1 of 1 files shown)


recena:
@reviewbybees Could you review these changes? /cc @christ66

recena:
In my local development environment are working fine.

christ66:
:bee: . I executed this pull request on my local machine and was unable to reproduce the failure.

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -1392,7 +1392,13 @@ else if (project.getLastBuild()!=null) {
 
         final Map<String,ISVNAuthenticationProvider> authProviders = new LinkedHashMap<String,
                 ISVNAuthenticationProvider>();
-        for (ModuleLocation loc: getLocations()) {
+
+        EnvVars env = null;
+        if (project.getLastCompletedBuild() != null) {
```
oleg-nenashev: Wrong construct, which will make the static analysis angry (e.g. cleanup of builds between two calls). You should assign the last completed build to a variable and then check&use it

recena: @oleg-nenashev Before to do any modifications, I need to understand what do you mean. What static analysis tool?

oleg-nenashev: Any.
Result of "project.getLastCompletedBuild()" may change between two calls => such null check does not guarantee the result

recena:
@oleg-nenashev Thanks for your comment. Reviewed.

oleg-nenashev:
:bee:

daniel-beck:
This change does not consider environment variables changed since the previous build. As replacement to #83 it is unsuitable. (And FWIW I consider reuse of previous builds' environments to be an antipattern.)

recena:
@daniel-beck I agree. But I didn't know how to get the environment vars (all of them). Any idea?

recena:
@daniel-beck I'm going to work in order to improve the current merged PR.

PR #128, opened 2015-07-19
[JENKINS-26458] Guard against a workspace format of 100.

christ66:
https://issues.jenkins-ci.org/browse/JENKINS-26458

There are several plugins which are using an older version of the svnkit library which still use the workspace format of `100`. Since previous versions of this plugin used 100 as the workspace format this will guard against only an invalid format of `100`.

### DIFF
--- a/src/main/java/hudson/scm/subversion/CheckoutUpdater.java
+++ b/src/main/java/hudson/scm/subversion/CheckoutUpdater.java
@@ -39,6 +39,7 @@ import org.tmatesoft.svn.core.SVNCancelException;
 import org.tmatesoft.svn.core.SVNDepth;
 import org.tmatesoft.svn.core.SVNException;
 import org.tmatesoft.svn.core.internal.util.DefaultSVNDebugLogger;
+import org.tmatesoft.svn.core.internal.wc17.db.ISVNWCDb;
 import org.tmatesoft.svn.core.internal.wc2.compat.SvnCodec;
 import org.tmatesoft.svn.core.wc.SVNRevision;
 import org.tmatesoft.svn.core.wc.SVNUpdateClient;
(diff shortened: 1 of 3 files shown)


recena:
@christ66 Is this PR ready to review by bees?

christ66:
@reviewbybees

daniel-beck:
Do we know that `100` is 1.7?

src/main/java/hudson/scm/subversion/CheckoutUpdater.java
```diff
@@ -109,6 +110,11 @@ public UpdateTask createTask() {
                     checkout.setIgnoreExternals(location.isIgnoreExternalsOption());
                     checkout.setExternalsHandler(SvnCodec.externalsHandler(svnuc.getExternalsHandler()));
 
+                    // Statement to guard against JENKINS-26458.
+                    if (SubversionWorkspaceSelector.workspaceFormat == 100) {
```
oleg-nenashev: I would prefer to see a constant defining such value

christ66:
@daniel-beck Yes. From the 2.4.5 release, [global.jelly](https://github.com/jenkinsci/subversion-plugin/blob/subversion-2.4.5/src/main/resources/hudson/scm/SubversionSCM/global.jelly#L33) and [SubversionWorkspaceSelector.java#L129](https://github.com/jenkinsci/subversion-plugin/blob/subversion-2.4.5/src/main/java/hudson/scm/SubversionWorkspaceSelector.java#L129) both use the workspace format of `100`.

jglick:
:bee:

recena:
:bee:

recena:
@reviewbybees done

PR #129, opened 2015-07-24
SVNRepository.getRepositoryRoot(false) can return null

jglick:
While working on Subversion-based integration tests for https://github.com/jenkinsci/workflow-plugin/pull/147 I found that branch indexing would fail with the following error:

```
Indexing started at …
Started
Opening conection to file:///…reporoot/
FATAL: Failed to index
java.lang.NullPointerException
	at hudson.scm.SubversionSCM$DescriptorImpl.getRelativePath(SubversionSCM.java:2375)
	at jenkins.scm.impl.subversion.SubversionSCMSource.retrieve(SubversionSCMSource.java:218)
	at jenkins.scm.api.SCMSource.fetch(SCMSource.java:141)
	at jenkins.branch.MultiBranchProject.populateBranchItemsFromSCM(MultiBranchProject.java:727)
	at jenkins.branch.MultiBranchProject.access$500(MultiBranchProject.java:160)
	at jenkins.branch.MultiBranchProject$BranchIndexing.run(MultiBranchProject.java:1873)
	at hudson.model.ResourceController.execute(ResourceController.java:98)
	at hudson.model.Executor.run(Executor.java:374)
Finished: FAILURE
```

Here `SVNRepository.getRepositoryRoot(false)` is being called, which does not guarantee a non-null result:

```java
public SVNURL getRepositoryRoot(boolean forceConnection) throws SVNException {
    if (forceConnection && myRepositoryRoot == null) {
        testConnection();
    }
    return myRepositoryRoot;
}
```

You need to pass `true` to force `testConnection()` to be called and thus `myRepositoryRoot` to be set.

@tfennelly claimed indexing worked for him, so there may be some specific circumstance in which this failure occurs. I suspect the difference is that I am testing against a local (`file`-protocol) repository, whereas when using HTTP(S) some other part of the code forces a connection to be made earlier. I did not try using `svnserve`.

I found three places calling `getRepositoryRoot`, all passing `false`, and all assuming that the result was nonnull. I fixed them all, though I can only verify the fix in the case mentioned in the stack trace: with that change my test passes. I believe the fix is safe, since either `myRepositoryRoot` was null, in which case the patch is needed to avoid an NPE, or it was not, in which case passing `true` has no effect.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -2372,7 +2372,7 @@ public class SubversionSCM extends SCM implements Serializable {
         }
 
         public static String getRelativePath(SVNURL repoURL, SVNRepository repository) throws SVNException {
-            String repoPath = repoURL.getPath().substring(repository.getRepositoryRoot(false).getPath().length());
+            String repoPath = repoURL.getPath().substring(repository.getRepositoryRoot(true).getPath().length());
             if(!repoPath.startsWith("/"))    repoPath="/"+repoPath;
             return repoPath;
         }
@@ -2747,7 +2747,7 @@ public class SubversionSCM extends SCM implements Serializable {
                     if (r.getRepositoryUUID(false) == null)
                         r.testConnection(); // make sure values are fetched
                     repositoryUUID = UUID.fromString(r.getRepositoryUUID(false));
-                    repositoryRoot = r.getRepositoryRoot(false);
+                    repositoryRoot = r.getRepositoryRoot(true);
                 }
             }
             return repositoryUUID;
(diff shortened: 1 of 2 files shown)


recena:
:bee: but I'd move the annotation.

recena:
@jglick I plan to cut a release (2.5.2) at the end of this month. Would it be great for you?

oleg-nenashev:
:bee: from the functional side.
Should we expect an impact on the performance due to sequential testConnection() calls in a same method?

jglick:
> We plan to cut a release (2.5.2) at the end of this month. Would it be great for you?

As noted in the PR description, this _probably_ does not matter for real users.

> Should we expect an impact on the performance due to sequential testConnection() calls in a same method?

I would not expect so, since the call is only made if the repository location is not already cached. As noted in the PR description, unless I am mistaken somehow, the patch should only change the behavior in cases where the current code would have thrown an NPE.

jglick:
Hmm, using a current Subversion plugin release I still get an NPE during branch indexing when I use a `svn://localhost/path/to/svnrepo/` URL from a Literate project after running `svnserve -d`. Whereas `https://svn.jenkins-ci.org/trunk/hudson/test-projects/trivial-ant/` works OK (does not throw an NPE, at least). And a demo project on GitHub accessed using Subversion, `https://github.com/jglick/literate-demo/`, gets indexed fine (the branch project build later fails to check out but this seems to be generally true of SVNKit and GitHub). So maybe it only works today when using an `https`-protocol URL?

PR #131, opened 2015-08-15
[JENKINS-29340] ISVNAuthentication provider did not provide credentials (part. II)

recena:
Downstream of https://github.com/jenkinsci/envinject-plugin/pull/56

This PR is related to https://github.com/jenkinsci/subversion-plugin/pull/127. @daniel-beck told [here](https://github.com/jenkinsci/subversion-plugin/pull/127#issuecomment-122901366) a side-effect and I've been working on a different approach.

Tickets involved: [JENKINS-29340](https://issues.jenkins-ci.org/browse/JENKINS-29340) and [JENKINS-23007](https://issues.jenkins-ci.org/browse/JENKINS-23007)

@reviewbybees

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -78,6 +78,7 @@ import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.WeakHashMap;
 
+import hudson.remoting.LocalChannel;
 import hudson.security.ACL;
 import hudson.util.ListBoxModel;
 import jenkins.model.Jenkins;
@@ -1332,11 +1333,9 @@ public class SubversionSCM extends SCM implements Serializable {
         final SVNRevisionState baseline;
         if (_baseline instanceof SVNRevisionState) {
             baseline = (SVNRevisionState)_baseline;
-        }
-        else if (project.getLastBuild()!=null) {
+        } else if (project.getLastBuild()!=null) {
             baseline = (SVNRevisionState)calcRevisionsFromBuild(project.getLastBuild(), launcher != null ? workspace : null, launcher, listener);
-        }
-        else {
+        } else {
             baseline = new SVNRevisionState(null);
         }
 
@@ -1354,9 +1353,8 @@ public class SubversionSCM extends SCM implements Serializable {
             }
             if (project instanceof AbstractProject && repositoryLocationsNoLongerExist(lastCompletedBuild, listener, env)) {
                 // Disable this project, see HUDSON-763
-                listener.getLogger().println(
-                        Messages.SubversionSCM_pollChanges_locationsNoLongerExist(project));
-                disableProject((AbstractProject) project, listener);              
+                listener.getLogger().println(Messages.SubversionSCM_pollChanges_locationsNoLongerExist(project));
+                disableProject((AbstractProject) project, listener);
                 return NO_CHANGES;
             }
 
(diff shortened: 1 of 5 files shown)


src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -1354,9 +1353,8 @@ else if (project.getLastBuild()!=null) {
             }
             if (project instanceof AbstractProject && repositoryLocationsNoLongerExist(lastCompletedBuild, listener, env)) {
                 // Disable this project, see HUDSON-763
-                listener.getLogger().println(
-                        Messages.SubversionSCM_pollChanges_locationsNoLongerExist(project));
-                disableProject((AbstractProject) project, listener);              
+                listener.getLogger().println(Messages.SubversionSCM_pollChanges_locationsNoLongerExist(project));
```
daniel-beck: Too many reformatting changes make the PR unnecessarily difficult to read.

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -1371,33 +1369,34 @@ else if (project.getLastBuild()!=null) {
...
-        final Map<String,ISVNAuthenticationProvider> authProviders = new LinkedHashMap<String,
-                ISVNAuthenticationProvider>();
+        final Map<String,ISVNAuthenticationProvider> authProviders = new LinkedHashMap<String, ISVNAuthenticationProvider>();
 
-        EnvVars env = null;
-        Run<?,?> run = project.getLastCompletedBuild();
-        if (run != null) {
-            env = run.getEnvironment(listener);
-        }
+        Node node = Jenkins.getInstance();
```
daniel-beck: Why is this always using the master node? You removed `isPollFromMaster()` which implies there are other possibilities.

daniel-beck:
Does this not fix JENKINS-23007 or why are you not referencing it?

src/main/java/hudson/scm/SubversionSCM.java
recena: @daniel-beck I agree. I'm trying to apply a coding style and I do modifications in each PRs.

src/main/java/hudson/scm/SubversionSCM.java
recena: @daniel-beck I'm going to investigate how I can get the right `Node`

recena:
@daniel-beck I have to verify [JENKINS-23007](https://issues.jenkins-ci.org/browse/JENKINS-23007). Inicially, this PR + https://github.com/jenkinsci/envinject-plugin/pull/56, resolve this bug [JENKINS-29340](https://issues.jenkins-ci.org/browse/JENKINS-29340)

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -1371,33 +1369,34 @@ else if (project.getLastBuild()!=null) {
...
-                ISVNAuthenticationProvider>();
+        final Map<String,ISVNAuthenticationProvider> authProviders = new LinkedHashMap<String, ISVNAuthenticationProvider>();
 
-        EnvVars env = null;
-        Run<?,?> run = project.getLastCompletedBuild();
-        if (run != null) {
-            env = run.getEnvironment(listener);
-        }
+        Node node = Jenkins.getInstance().getNode(nodeName);
+
```
recena: @daniel-beck What do you think? Better?

daniel-beck: I don't think this is correct. `getNode(String)` can never return the master node, so if `isPollFromMaster()` or the last build was on the master node, this should fail. Did you test this at all?

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -1371,34 +1369,40 @@ else if (project.getLastBuild()!=null) {
...
 
-        final Map<String,ISVNAuthenticationProvider> authProviders = new LinkedHashMap<String,
-                ISVNAuthenticationProvider>();
+        final Map<String,ISVNAuthenticationProvider> authProviders = new LinkedHashMap<String, ISVNAuthenticationProvider>();
 
-        EnvVars env = null;
-        Run<?,?> run = project.getLastCompletedBuild();
-        if (run != null) {
-            env = run.getEnvironment(listener);
+        Node node;
```
daniel-beck: It looks like you added more code and complexity above this line [compared to 2.5.1](https://github.com/jenkinsci/subversion-plugin/compare/subversion-2.5.1...recena:JENKINS-29340#diff-d1e0e212be7f244b33835b2b65a19f62L1381) (new line 1378...new line 1389) without actually changing the behavior. Or was the behavior changed in some way that I'm not aware of?

recena: @daniel-beck I only tried to do the code more readable, at least, in my opinion. I'll change it if you consider that I'm wrong.

src/test/java/hudson/scm/SubversionSCMTest.java
```diff
@@ -61,7 +61,6 @@
 import org.tmatesoft.svn.core.wc.SVNCommitClient;
 import org.tmatesoft.svn.core.wc.SVNRevision;
 import org.tmatesoft.svn.core.wc.SVNStatus;
-import org.tmatesoft.svn.core.wc.SVNWCUtil;
```
christ66: Unnecessary change

recena: @christ66 But it is an unnecessary import. What do you propose? A dedicated PR for that???

christ66: It adds unnecessary history to the diff. If I want to view the changes made for JENKINS-29340, I should only see the _necessary_ changes. The removal of an import in `SubversionSCMTest` has nothing to do with the changes for JENKINS-29340. We should create a new PR that does this type of cleanup. Comment is more about future pull requests.

recena: I disagree but I'll do it.

recena: @christ66 Done.

oleg-nenashev:
If I understand correctly, a new test is the only real change. 
:bee: for the test, :ant: for the unnecessary and confusing refactoring.

src/test/java/hudson/scm/SubversionEnvInjectTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+        project.setScm(new SubversionSCM(REPO_URL));
+
+        TaskListener listener = jenkins.createTaskListener();
+        PollingResult poll = project.poll(listener);
+        assertTrue(poll.hasChanges());
+
+        jenkins.assertBuildStatusSuccess(project.scheduleBuild2(0).get());
+    }
+}
```
tfennelly: Hey Manuel, would you be able to add some comments to the test to explain how it is testing this? E.g. of what I am wondering about is what is the importance of `poll.hasChanges()` being `true` and what was done earlier in the test that should result in it being `true`?

tfennelly:
:bee: as far as I understand it.

recena:
@oleg-nenashev This PR can not be reviewed only seeing the source code changes, the context is very important:

* There is a PR in EnvInject Plugin involved.
* There is a merged PR related

recena:
@tfennelly I've added some comments to test. Do you agree?

recena:
@reviewbybees done

PR #141, opened 2015-10-27
[JENKINS-30196] SVN Exception : svn: E210004: Malformed network data

recena:
[JENKINS-30196](https://issues.jenkins-ci.org/browse/JENKINS-30196)

PR #142, opened 2015-10-30
[FIXED JENKINS-31192] Refactor HTTPS SSL client certificate support

ymartin59:
To allow certificate tranfer as byte array between master and slave,
RemoteSVNSSLAuthentication extends SVNKit original SVNSSLAuthentication
and generates a temporary PKCS#12 file when used for authentication.

Resolves: JENKINS-31192

### DIFF
--- a/src/main/java/hudson/scm/CredentialsSVNAuthenticationProviderImpl.java
+++ b/src/main/java/hudson/scm/CredentialsSVNAuthenticationProviderImpl.java
@@ -24,6 +24,8 @@ import org.tmatesoft.svn.core.auth.*;
 
 import javax.security.auth.DestroyFailedException;
 import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.Serializable;
 import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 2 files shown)


recena:
@ymartin59 Reviewing...

recena:
@ymartin59 Have you considered to do these changes in SVNKit library?

ymartin59:
@recena Hello. No (almost). As I know the plugin was using a modified version of SVNKit until recently. So I guess that plugin maintainer already tried to get "byte array certificate support" included in SVNKit (and probably get a "no" as answer. But that it is clear it would be far better that creating temporary file on disk - even if my first look at SVNKit sources did not point me to the "right" place where certificate reading occurs.

recena:
@ymartin59 Well, I'm trying to change this. I want to use the official release here https://github.com/jenkinsci/subversion-plugin/pull/141

ymartin59:
@recena I have just tested with #141 included. This PR for SSL PKCS#12 certification still works well.

I will design a patch for SVNKit and see how it is welcome.

ymartin59:
@recena Sorry I probably misunderstood your point. I just saw svnkit 1.8.11 has support for certificate as byte array. So I will propose a PR based on svnkit 1.8.11.

ymartin59:
@recena Give up this PR. Please review #145

PR #144, opened 2015-11-04
[FIXED JENKINS-31385] Additional logs when using svn:externals is used and SVNCancelException is thrown

fbelzunc:
[JENKINS-31385](https://issues.jenkins-ci.org/browse/JENKINS-31385)

If `svn:externals` is used and the user has not configured *Additional Credentials* then `SVNCancelException` is thrown. A lot of times subversion users don't know they need to configure the external credentials on `Additional Credentials` section, so it might be a good idea to catch this exception and provide some user-friendly information.

@reviewbybees @recena @christ66


recena:

PR #148, opened 2015-12-11
[JENKINS-14155] Automatically select the first option of tags listing if default value is not defined

recena:
[JENKINS-14155](https://issues.jenkins-ci.org/browse/JENKINS-14155) ([JENKINS-19560](https://issues.jenkins-ci.org/browse/JENKINS-19560) duplicated)

@reviewbybees


oleg-nenashev:
I'm not a big fan of such change. It's not "the first in the list", it's "first in the list of tags without the context".  So it's not very predictable

recena:
@oleg-nenashev Well, this PR provides some case-uses demanded by users. My concern here is about the performance. What happens if the list is very large? :tired_face:

recena:

PR #150, opened 2015-12-22
findbugs plugin has been updated. 2 errors were solved

recena:
@reviewbybees

### DIFF
--- a/src/test/java/hudson/scm/SubversionSCMTest.java
+++ b/src/test/java/hudson/scm/SubversionSCMTest.java
@@ -97,8 +97,8 @@ public class SubversionSCMTest extends AbstractSubversionTest {
         super.setUp();
         repo = SVNURL.parseURIDecoded("http://subversion.tigris.org/svn/subclipse");
 
-        // during the test, don't pollute the global authentication cache.
-        SubversionSCM.CONFIG_DIR = createTmpDir().getAbsolutePath();
+        // during the test, don't pollute the user's configuration (esp. authentication cache).
+        System.setProperty(SubversionSCM.class.getName() + ".configDir", createTmpDir().getAbsolutePath());
     }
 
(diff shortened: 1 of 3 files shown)


andresrc:
:bee:

src/test/java/hudson/scm/SubversionSCMTest.java
```diff
@@ -97,8 +97,8 @@ protected void setUp() throws Exception {
         super.setUp();
         repo = SVNURL.parseURIDecoded("http://subversion.tigris.org/svn/subclipse");
 
-        // during the test, don't pollute the global authentication cache.
-        SubversionSCM.CONFIG_DIR = createTmpDir().getAbsolutePath();
+        // during the test, don't pollute the user's configuration (esp. authentication cache).
+        System.setProperty(SubversionSCM.class.getName() + ".configDir", createTmpDir().getAbsolutePath());
```
amuniz: Is this already visible on `SubversionSCM ` class loading?

amuniz:
:bee:

src/test/java/hudson/scm/SubversionSCMTest.java
recena: Really, in this point you define the environment variable, only that. When `SubversionSCM` is instantiated, it hopes that `SubversionSCM.CONFIG_DIR` have this value for the tests. [Reference](https://github.com/jenkinsci/subversion-plugin/blob/master/src/main/java/hudson/scm/SubversionSCM.java#L3149) happens.

recena:
@reviewbybees done

src/test/java/hudson/scm/SubversionSCMTest.java
recena: And yes, `SubversionSCM` class is loaded in this poing.

PR #153, opened 2015-12-30
[JENKINS-16711] Encode the URLs when are shown in the logger

recena:
We want to avoid this:

jenkins-16711

@reviewbybees


recena:

PR #157, opened 2016-02-09
Use the lastest changed tag as default

MichaWiedenmann:
In a recent commit (5012326e8dc523e5e7706839b7853a86b1f3dce5 - [JENKINS-14155] Automatically select the first option of tags listing…"), support was added to select the latest changed tag as default value. For this to work, the repository URL (of the subversion tag parameters) must point to the tags directory (e.g. foo/tags). If you want to also include trunk and branches (i.e. setting the repository URL to foo/), you cannot benefit from the "use the latest tag" feature.

This branch extends the recent functionality, such that one can use a repository URL pointing to the top level directory (foo) while still using the latest changed tag from the tags sub directory as default.

To this end a new checkbox "Use latest tag" is added below the "Default value" text box.


kuisathaverat:
Closed and Open to relaunch CI

src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
```diff
@@ -98,29 +98,31 @@
...
   @Deprecated
-  public ListSubversionTagsParameterDefinition(String name, String tagsDir, String tagsFilter, String defaultValue, String maxTags, boolean reverseByDate, boolean reverseByName, String uuid, String credentialsId) {
-      this(name, tagsDir, credentialsId, tagsFilter, defaultValue, maxTags, reverseByDate, reverseByName);
+  public ListSubversionTagsParameterDefinition(String name, String tagsDir, String tagsFilter, String defaultValue, boolean useLatestTag, String maxTags, boolean reverseByDate, boolean reverseByName, String uuid, String credentialsId) {
+      this(name, tagsDir, credentialsId, tagsFilter, defaultValue, useLatestTag, maxTags, reverseByDate, reverseByName);
   }
 
   @DataBoundConstructor
-  public ListSubversionTagsParameterDefinition(String name, String tagsDir, String credentialsId, String tagsFilter, String defaultValue, String maxTags, boolean reverseByDate, boolean reverseByName) {
+  public ListSubversionTagsParameterDefinition(String name, String tagsDir, String credentialsId, String tagsFilter, String defaultValue, boolean useLatestTag, String maxTags, boolean reverseByDate, boolean reverseByName) {
```
Wadeck: I dunno exactly the guideline for this plugin, but from my PoV it's better to use `@DataBoundSetter` when you add field instead of putting that in the constructor.

src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
```diff
@@ -98,29 +98,31 @@
...
+  private final boolean useLatestTag;
   private final String maxTags;
   private static final String SVN_BRANCHES = "branches";
   private static final String SVN_TAGS = "tags";
   private static final String SVN_TRUNK = "trunk";
   
   @Deprecated
-  public ListSubversionTagsParameterDefinition(String name, String tagsDir, String tagsFilter, String defaultValue, String maxTags, boolean reverseByDate, boolean reverseByName, String uuid) {
-    this(name, tagsDir, null, tagsFilter, defaultValue, maxTags, reverseByDate, reverseByName);
+  public ListSubversionTagsParameterDefinition(String name, String tagsDir, String tagsFilter, String defaultValue, boolean useLatestTag, String maxTags, boolean reverseByDate, boolean reverseByName, String uuid) {
```
Wadeck: Those additions will result in breaking the binary compatibility. Is it required to pass by the constructor ? I think a simple use of a setter / putting a default value would be sufficient.

src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
```diff
@@ -173,6 +181,10 @@ public DescriptorImpl getDescriptor() {
     return (DescriptorImpl) super.getDescriptor();
   }
 
+  @Nonnull public List<String> getTags(@Nullable Job context) {
+      return getTags(context, "", isReverseByDate()); 
+  }
```
Wadeck: 👍 to think about compatibility

PR #158, opened 2016-03-25
Fixes to SVNRepositoryView cache

jglick:
A user of `SubversionSCMSource` reported hangs in branch indexing. Thread dumps showed thread activity in locking code, not in SVNKit.

@reviewbybees esp. @recena & @stephenc

### DIFF
--- a/src/main/java/jenkins/scm/impl/subversion/SVNRepositoryView.java
+++ b/src/main/java/jenkins/scm/impl/subversion/SVNRepositoryView.java
@@ -25,6 +25,7 @@ package jenkins.scm.impl.subversion;
 
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.scm.*;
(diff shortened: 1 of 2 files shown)


src/main/java/jenkins/scm/impl/subversion/SVNRepositoryView.java
```diff
@@ -84,7 +89,10 @@ public int getReadTimeout(SVNRepository repository) {
         repository.setTunnelProvider(SVNWCUtil.createDefaultOptions(true));
         repository.setAuthenticationManager(sam);
         try {
-            uuid = repository.getRepositoryUUID(false);
+            uuid = repository.getRepositoryUUID(true);
+            if (uuid == null) { // TODO is this even possible? Javadoc is unclear.
```
jglick: Noticed when running `SCMBinderTest.exactRevisionSubversion` that `uuid` was null, which certainly seemed wrong in this context—that would mean every source would be sharing a single cache! Not sure if this problem is limited to local file connections (as opposed to HTTPS), but seems best to require a defined UUID.

andresrc:
:bee:

recena:
@jglick LGTM :+1:

PR #159, opened 2016-05-20
Fix bug where having branch wildcards at different depths would cause ArrayBoundExceptions

frankier:
An example use case of this is the Jenkins Multibranch Pipeline to try and find stuff to build in a chaotic monorepo. I set "Include branches" to: trunk,branches/*,branches/*/*,branches/*/*/*,branches/*/*/*/*,tags/* and got an indexing exception. This patch seems to fix it.


recena:
@frankier Is there any Jira issue for that?

oleg-nenashev:
Without knowing the stacktrace, it's hard to decide. The patch seems to be valid, but I agree with previous reviewers

PR #140, opened 2015-10-24
[WiP] [JENKINS-31090] Not compiling against Jenkins 1.580+

recena:
[JENKINS-31090](https://issues.jenkins-ci.org/browse/JENKINS-31090)

@kohsuke, @daniel-beck, your help here will be great!

PR #161, opened 2016-06-01
[JENKINS-35247] Moved SubversionStep to subversion-plugin

jglick:
[JENKINS-35247](https://issues.jenkins-ci.org/browse/JENKINS-35247)

Probably supersedes #140.

@reviewbybees


jglick:
Probably going to fail with a cyclic dependency error for the same reason as https://github.com/jenkinsci/git-plugin/pull/406.

src/main/java/jenkins/scm/impl/subversion/SubversionStep.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2014 Jesse Glick.
```
recena: ?¿

src/main/resources/jenkins/scm/impl/subversion/SubversionStep/config.jelly
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+The MIT License
+
+Copyright 2014 Jesse Glick.
```
recena: Ditto

recena:
@jglick Thanks for this special contribution. Let me review it.

src/main/java/jenkins/scm/impl/subversion/SubversionStep.java
jglick: Just moving the original source file.

jglick:
I am still checking whether I have gotten all the tests to pass. Most of the issues seem to have nothing to do with this patch per se, the plugin was just broken in trunk.

jglick:
> the plugin was just broken in trunk

…or needed updates to build against the newest parent POM and an LTS baseline.

jglick:
BTW I get some local failures in `testPinnedExternals` & `testDepthOptions` running the whole build which do not reproduce when running individual test cases. I suspect the port allocation logic in `runSvnServe` is wrong.

jglick:
> I suspect the port allocation logic in `runSvnServe` is wrong.

Confirmed, broken in #94.

jglick:
At any rate, the CI build seems clean now. Will proceed to release dependencies.

src/test/java/hudson/scm/SVNWorkingCopyTest.java
```diff
@@ -40,10 +40,12 @@
 import org.tmatesoft.svn.core.wc2.SvnGetStatus;
 
 import java.io.File;
+import org.junit.Ignore;
 
 /**
  * @author schristou88
  */
+@Ignore("TODO org.tmatesoft.svn.core.SVNException: svn: E175002: PROPFIND of '/trunk/hudson/test-projects/trivial-ant': 405 Method Not Allowed (https://svn.jenkins-ci.org)")
```
christ66: I think you're getting this error because https://svn.jenkins-ci.org doesn't exist anymore. Probably need to point it to a new subversion repository.

jglick: Yes, has been discussed on jenkinsci-dev. For purposes of this PR I just wanted to clear the standing test errors.

recena: @christ66 A new environment for Subversion server is underway, meantime I agree with @jglick's proposal in this PR.

jglick:
@recena ready for review I think.

christ66:
🐝

jglick:
`git` 2.5.0 was released, so just waiting on this PR to be merged and released before I can merge the downstream PR.

recena:
@jglick Reviewing the changes in local. Some tests are failing.

recena:
@jglick As I commented you offline, this testing approach (based on Subversion Client CLI) generates a dependency with the testing environment:

```
Subversion is open source software, see http://subversion.apache.org/
The following repository back-end (FS) modules are available:
* fs_fs : Module for working with a plain file (FSFS) repository.
svnadmin: invalid option: --compatible-version=1.5
Type 'svnadmin help' for usage.
```

jglick:
I see, you are referring to `SubversionSampleRepoRule.svn`. Yes, this could probably be migrated to SVNKit to make tests less platform-dependent. It works OK in our current CI environment and on my machine but indeed it could fail in other circumstances.

recena:
@jglick Exactly.

jglick:
Though [this code](https://github.com/jenkinsci/subversion-plugin/blob/fe86c32c504bc81fb9aefe68f365c21c3cd16075/src/test/java/hudson/scm/AbstractSubversionTest.java#L60-L65) shows that you were _already_ running at least `svnserve` as a forked processes for tests. Is that somehow different than running `svn` or `svnadmin`?

Also [this code](https://github.com/jenkinsci/subversion-plugin/blob/fe86c32c504bc81fb9aefe68f365c21c3cd16075/src/test/java/jenkins/scm/impl/subversion/AbstractSampleRepoRule.java#L57) should have been making the tests merely be skipped, not failing. Did that not work for you?

recena:
@jglick All of this was *there*, my comment was only a way to say: _we should not continue_ doing that. At any rate, the PR was merged and I appreciate your contribution.

jglick:
@recena stay tuned…

jglick:
Seems that `svnadmin --compatible-version` is 1.8+ only.

PR #165, opened 2016-08-04
[JENKINS-37184] [JENKINS-5124] Log the exception that is thrown if Jenksins can't get info() on a subversion repository using a tunnel

openbrian:
Log the exception that is thrown if Jenksins can't get info() on a subversion repository using a tunnel.  When the repository factory creates the repository, set the tunnel provider based on default svn options.

### DIFF
--- a/src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
+++ b/src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
@@ -179,6 +179,7 @@ public class ListSubversionTagsParameterDefinition extends ParameterDefinition {
       SVNURL repoURL = SVNURL.parseURIDecoded(getTagsDir());
 
       SVNRepository repo = SVNRepositoryFactory.create(repoURL);
+      repo.setTunnelProvider( SubversionSCM.createDefaultSVNOptions() );
       repo.setAuthenticationManager(authManager);
       SVNLogClient logClient = new SVNLogClient(authManager, null);
       
@@ -255,6 +256,7 @@ public class ListSubversionTagsParameterDefinition extends ParameterDefinition {
         return true;
       }
     } catch (SVNException e) {
+      LOGGER.log(Level.SEVERE, "An SVN exception occurred", e);
       return false;
     }
     return false;
--- a/src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
+++ b/src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
@@ -179,7 +179,7 @@ public class ListSubversionTagsParameterDefinition extends ParameterDefinition {
       SVNURL repoURL = SVNURL.parseURIDecoded(getTagsDir());
 
       SVNRepository repo = SVNRepositoryFactory.create(repoURL);
-      repo.setTunnelProvider( SubversionSCM.createDefaultSVNOptions() );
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
```diff
@@ -255,6 +256,7 @@ private boolean isSVNRepositoryProjectRoot(SVNRepository repo) {
         return true;
       }
     } catch (SVNException e) {
+      LOGGER.log(Level.SEVERE, "An SVN exception occurred", e);
```
recena: The message does not provide any value

src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
```diff
@@ -179,6 +179,7 @@ public DescriptorImpl getDescriptor() {
       SVNURL repoURL = SVNURL.parseURIDecoded(getTagsDir());
 
       SVNRepository repo = SVNRepositoryFactory.create(repoURL);
+      repo.setTunnelProvider( SubversionSCM.createDefaultSVNOptions() );
```
recena: Please, `repo.setTunnelProvider(SubversionSCM.createDefaultSVNOptions());`

src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
openbrian: It's doesn't?  It tells a sysadmin that they have a misconfigured repository line when they specify svn+foo://domain/path and foo is not a configured tunnel in .subversion/config.  It's much better then throwing that exception on the floor.  It's the only way I would know how to fix the bug.

src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
openbrian: I'm on it.

openbrian:
I believe this patch is ready to be merged.  If not, please provide feedback.

openbrian:
Any updates on this PR?

src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
```diff
@@ -179,6 +179,7 @@ public DescriptorImpl getDescriptor() {
       SVNURL repoURL = SVNURL.parseURIDecoded(getTagsDir());
 
       SVNRepository repo = SVNRepositoryFactory.create(repoURL);
+      repo.setTunnelProvider(SubversionSCM.createDefaultSVNOptions());
```
jglick: This is not about logging, is it?

jglick (changes requested):
merge conflicts

openbrian:
Well it's been about a year, so I don't remember.  I'll review and update the ticket.

I can tell you, the efect of this omission is, under some conditions, the list of known braches and tags will not populate a drop down select box. I believe it was like Build Parameterized or some form like that.

The root cause of the issue was not being logged.  We didn't know what to fix. The feature was unusable.

I don't recall what the root cause was, but I feel it was related to the fact that we use ssh+svn and our ssh uses Kerberos for authentication.

Eccenux:
Merge conflict; possibly not a problem any more.

PR #167, opened 2016-08-20
Should be using Job instead of Item

christ66:
Setting the `doFillCredentialsIdItems` to use `Item` instead of `Job` causes credentials defined at the folder level to not properly be made available with certain types of projects.

Similar to the mercurial plugin :https://github.com/jenkinsci/mercurial-plugin/blob/41fda47ac7749cc55173cb91e0dc5d953ddf90df/src/main/java/hudson/plugins/mercurial/MercurialSCM.java#L1048.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -3012,7 +3012,7 @@ public class SubversionSCM extends SCM implements Serializable {
                 return null;
             }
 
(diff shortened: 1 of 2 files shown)


Evildethow:
LGTM 🐝

oleg-nenashev:
Formally it breaks the binary compatibility.
But these stapler methods are not really supposed to be used outside IMHO.
I would just add <code>@Restricted</code> annotation

jglick:
:-1:, see other PR.

PR #170, opened 2016-10-11
Upgrade to svnkit 1.8.14.

mdiers:
SVNKit 1.8.14, released 10 Oct 2016, provides [a number of bug fixes](https://svn.svnkit.com/repos/svnkit/tags/1.8.14/CHANGES.txt).
It also addresses [SVNKIT-672](https://issues.tmatesoft.com/issue/SVNKIT-672), vastly improving performance on 32-bit Linux.

PR #173, opened 2016-10-19
Fixing repository declarations

jglick:
Jenkins plugins should specify the Jenkins repository to make sure they can be built without a custom `~/.m2/settings.xml`, so part of #172 was wrong. (Inheriting `distributionManagement` is correct.)
#170 was not wrong, but the original use of a custom repository here was. Fixed by manually uploading the [third-party library in question](https://repo.jenkins-ci.org/releases/com/trilead/trilead-ssh2/1.0.0-build221/). That also allows other plugins to compile against `subversion` 2.7.1 without hassle.

@reviewbybees esp. @recena, @daniel-beck

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,10 +70,16 @@ THE SOFTWARE.
 
   <repositories>
     <repository>
-      <id>mavenTMateSoftReleasesRepository</id>
-      <url>http://maven.tmatesoft.com/content/repositories/releases/</url>
+      <id>repo.jenkins-ci.org</id>
+      <url>https://repo.jenkins-ci.org/public/</url>
     </repository>
   </repositories>
+  <pluginRepositories>
+    <pluginRepository>
+      <id>repo.jenkins-ci.org</id>
+      <url>https://repo.jenkins-ci.org/public/</url>
+    </pluginRepository>
+  </pluginRepositories>
 
   <dependencies>
     <dependency>


mdiers:
Thanks. I was blissfully unaware of this particular complication when testing the #170 change.

pom.xml
```diff
@@ -70,10 +70,16 @@ THE SOFTWARE.
...
+      <id>repo.jenkins-ci.org</id>
+      <url>https://repo.jenkins-ci.org/public/</url>
     </repository>
   </repositories>
+  <pluginRepositories>
+    <pluginRepository>
+      <id>repo.jenkins-ci.org</id>
+      <url>https://repo.jenkins-ci.org/public/</url>
+    </pluginRepository>
+  </pluginRepositories>
```
recena: @jglick But, this configuration is defined (inherited) in parent [POM](https://github.com/jenkinsci/plugin-pom/blob/master/pom.xml#L711), right?

jglick: It is, but before the parent POM can even be applied, the `packaging` must be resolved…and `maven-hpi-plugin` is not in Central. Therefore if you remove this `pluginRepository`, the build will fail in a clean environment. (Of course the `repository` is also needed, else the parent POM cannot even be loaded.)

abayer:
:bee:

jglick:
@reviewbybees done

xylifyx:
A better solution would be to ask maintainers of org.jenkins-ci.plugins:plugin and org.jenkins-ci.tools:maven-hpi-plugin to release to maven central, then the declarations are not needed.

jglick:
> release to maven central

Ideally yes. There are fairly strict requirements on Central publishing, though, and it would take significant effort to ensure that we meet all the conditions. Anyway a topic for the dev list.

PR #174, opened 2016-11-05
[JENKINS-36521] Print raw (not HTML escaped) commit messages

ljader:
Commit decorators add clicable <a> links to issue IDs on build summary page,
but the security mechanism preventing XSS in .jelly,
escapes the HTML tags for .jelly files with:
jelly escape-by-default='true'

After the change, annotated commit messages are printed raw,
without HTML escaping.

Used method will be consistent with change in hudson/scm/SCM/project-changes.jelly
introduced in https://github.com/jenkinsci/jenkins/commit/41ab84fe0a1512fe52347d55fb58445174636896

Additional details:
https://wiki.jenkins-ci.org/display/JENKINS/Jelly+and+XSS+prevention
https://issues.jenkins-ci.org/browse/JENKINS-5135

### DIFF
--- a/src/main/resources/hudson/scm/SubversionChangeLogSet/digest.jelly
+++ b/src/main/resources/hudson/scm/SubversionChangeLogSet/digest.jelly
@@ -60,7 +60,7 @@ THE SOFTWARE.
       <ol>
         <j:forEach var="cs" items="${it.logs}" varStatus="loop">
           <li>
-            ${cs.msgAnnotated}
+            <j:out value="${cs.msgAnnotated}"/>
             (<a href="${changesBaseUrl}changes#detail${loop.index}">${%detail}</a>
 
             <j:set var="cslink" value="${browser.getChangeSetLink(cs)}"/>
--- a/src/main/resources/hudson/scm/SubversionChangeLogSet/index.jelly
+++ b/src/main/resources/hudson/scm/SubversionChangeLogSet/index.jelly
@@ -44,7 +44,7 @@ THE SOFTWARE.
               <a href="${browser.getChangeSetLink(cs)}">${cs.revision}</a>
               by <a href="${rootURL}/${cs.author.url}/">${cs.author}</a>:
             </b><br/>
-            ${cs.msgAnnotated}
+            <j:out value="${cs.msgAnnotated}"/>
           </div>
         </td>
       </tr>


jasek:
Works great. Thanks!

t0r0X:
:+1:  thanks!

ljader:
Can a svn-plugin maintainer review this?

CC @jenkinsci/code-reviewers

michaelpaesold:
:thumbsup:
Looks good, just wanted to propose the same fix and then found your PR.
Please merge!

ljader:
Hi @recena , can I merge this PR myself?

I'm regular member of jenkinsci GitHub group.

CC @jenkinsci/code-reviewers

ljader:
Hi again @recena and @jenkinsci/code-reviewers , FYI I will merge this PR myself this week.

michaelpaesold:
@oleg-nenashev Please check this again, your analysis is not correct as far as I can see. As I already pointed out below, the [getMsgAnnotated](https://github.com/jenkinsci/jenkins/blob/08def67a18eee51de9f3f99bc2a792fee1c160e0/core/src/main/java/hudson/scm/ChangeLogSet.java#L248) method returns output as marked up HTML, and itself escapes the commit message before returning it (it also replaces newlines with `<br>`, which are now incorrectly rendered as `&lt;br&gt;`).

It does **not** allow unescaped HTML from the commit message. What it does, is allow **other plugins**, such as the JIRA plugin, to add HTML markup.

I have also tested the PR snapshot build, for me e.g. `<b>text</b>` in the commit message is actually rendered as `&lt;b&gt;text&lt;/b&gt;` correctly, while formatting done by the JIRA plugin is rendered as HTML.

Please also compare the Git plugin, which does this exactly as proposed here.

daniel-beck:
@oleg-nenashev @ljader  I'll investigate this.

daniel-beck:
`ChangeLogSet` (in core) provides the `getMsgAnnotated()` method relevant here. It wraps the raw `getMsg()` in a `MarkupText`, whose `toString(boolean)` methods always escape HTML regardless of argument: https://github.com/jenkinsci/jenkins/blob/e7e51eeb62f1f235933740f02673fdc3c13102f3/core/src/main/java/hudson/MarkupText.java#L298...L321

Tests in https://github.com/jenkinsci/jenkins/blob/e7e51eeb62f1f235933740f02673fdc3c13102f3/core/src/test/java/hudson/MarkupTextTest.java ensure this behavior.

As `getMsgAnnotated()` escapes already, it's not relying on Stapler to ensure safe output.

So AFAICT security does not suffer from merging this.

CC @oleg-nenashev @michaelpaesold @ljader

oleg-nenashev:
Agreed

kassovix:
Any idea when this could be release ?

oleg-nenashev:
CC @recena . Do you still maintain the plugin? Or maybe @jglick ?

PR #176, opened 2017-01-07
add an svninfo step that can provide infos for a wc path

tobiasbaum:
Hi,

I implemented an "svninfo" step that provides information like the current revision for a working copy path. If you want to, you can include it in the official release.

Background: We're currently porting our builds to pipeline and miss the SVN_REVISION environment variable. As a workaround, we first tried to shell out and use a native svn to get the info. That works, but the dependency on a native svn (possibly different from the one Jenkins uses) is a potential maintenance problem. I looked some time for another solution but then went ahead implementing it.

I added some more of the provided infos to the returned map. I selected ones I thought could be interesting, but the selection is nevertheless rather arbitrary. It should be easy to extend.

Best regards,
   Tobias

### DIFF
--- a/src/main/java/jenkins/scm/impl/subversion/SvninfoStep.java
+++ b/src/main/java/jenkins/scm/impl/subversion/SvninfoStep.java
@@ -24,6 +24,7 @@
 package jenkins.scm.impl.subversion;
 
 import java.io.File;
+import java.io.IOException;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


src/main/java/jenkins/scm/impl/subversion/SvninfoStep.java
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        @Override
+        public Set<Class<?>> getRequiredContext() {
+            return Collections.<Class<?>>singleton(FilePath.class);
+        }
+
+    }
+
+    public static final class Execution extends AbstractSynchronousNonBlockingStepExecution<Map<String, String>> {
```
jglick: Update dep and use `SynchronousNonBlockingStepExecution`. Also need not be `public`.

src/main/java/jenkins/scm/impl/subversion/SvninfoStep.java
```diff
@@ -0,0 +1,143 @@
...
+        private static final long serialVersionUID = 1L;
+
+    }
+
+    private static final class SvnInfoFileCallable implements FileCallable<Map<String, String>> {
+
+        private static final long serialVersionUID = 1456116737119973646L;
+
+        @Override
+        public void checkRoles(final RoleChecker checker) throws SecurityException {
```
jglick: No. Use `MasterToSlaveFileCallable`.

tobiasbaum:
Updating the dependencies does not seem to be that straightforward, at least some tests fail now (ExtensionListListener cannot be found and the test setup does not work anymore). Perhaps I'll have some time to look at it in more depth in the next days.

Solving JENKINS-26100 would definitely be nice. I used the current solution mainly because it seemed easier to me. Also having an svninfo step is more powerful than just returning the revision, so I think they don't exclude each other.

jglick:
I think this is superseded by #182.

PR #177, opened 2017-01-31
[JENKINS-32167] Print log messages about which credentials, if any, are being associated with a given realm

jglick:
A quick attempt to see if enhanced logging makes [JENKINS-32167](https://issues.jenkins-ci.org/browse/JENKINS-32167) any clearer.

@reviewbybees esp. @stephenc

### DIFF
--- a/src/main/java/hudson/scm/CredentialsSVNAuthenticationProviderImpl.java
+++ b/src/main/java/hudson/scm/CredentialsSVNAuthenticationProviderImpl.java
@@ -323,7 +323,12 @@ public class CredentialsSVNAuthenticationProviderImpl implements ISVNAuthenticat
                 l.getLogger().printf("Found credentials %s in realm ‘%s’%n", CredentialsNameProvider.name(c), realm);
             } else {
                 c = defaultCredentials;
-                l.getLogger().printf("No credentials found for realm ‘%s’ among %s; falling back to %s%n", realm, credentialsByRealm.keySet(), defaultCredentials != null ? CredentialsNameProvider.name(defaultCredentials) : "<none>");
+                String name = c != null ? CredentialsNameProvider.name(c) : "<none>";
+                if (credentialsByRealm.isEmpty()) {
+                    l.getLogger().printf("Using sole credentials %s in realm ‘%s’%n", name, realm);
+                } else {
+                    l.getLogger().printf("No credentials found for realm ‘%s’ among %s; falling back to %s%n", realm, credentialsByRealm.keySet(), name);
+                }
             }
             if (c instanceof CertificateCredentials) {
                 return new SVNCertificateAuthenticationBuilder((CertificateCredentials) c);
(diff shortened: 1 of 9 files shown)


src/main/java/hudson/scm/CredentialsSVNAuthenticationProviderImpl.java
```diff
@@ -59,16 +61,23 @@
 
     public CredentialsSVNAuthenticationProviderImpl(Credentials credentials) {
         this.provider =
-                new RemotableSVNAuthenticationBuilderProvider(credentials, Collections.<String, Credentials>emptyMap());
+                new RemotableSVNAuthenticationBuilderProvider(credentials, Collections.<String, Credentials>emptyMap(), /* TODO */ TaskListener.NULL);
```
oleg-nenashev: Better to use the `this()` invocation of the constructor below

src/main/java/hudson/scm/CredentialsSVNAuthenticationProviderImpl.java
```diff
@@ -270,11 +292,20 @@ public void reset() {
 
         private final Credentials defaultCredentials;
         private final Map<String, Credentials> credentialsByRealm;
+        private final TaskListener listener;
```
oleg-nenashev: Add the CheckForNull annotation

src/main/java/hudson/scm/CredentialsSVNAuthenticationProviderImpl.java
```diff
@@ -270,11 +292,20 @@ public void reset() {
...
 
         public RemotableSVNAuthenticationBuilderProvider(Credentials defaultCredentials,
-                                                         Map<String, Credentials> credentialsByRealm) {
+                                                         Map<String, Credentials> credentialsByRealm,
+                                                         TaskListener listener) {
             this.defaultCredentials = defaultCredentials;
             this.credentialsByRealm = credentialsByRealm;
+            this.listener = listener == TaskListener.NULL ? null : listener;
```
oleg-nenashev: Why would we need it? One can copy-paste his own equivalent of `TaskListener.NULL`, which will have a different behavior then

jglick: > Why would we need it?

Because `NULL` is not `Serializable` I am afraid.



> One can copy-paste his own equivalent of `TaskListener.NULL`, which will have a different behavior then

Assuming it is `Serializable`, it should have the same behavior—nothing will be printed.

jglick:
Unfortunately no existing tests create an `AdditionalCredentials`, so it is not trivial to see the effect. (Some tests use externals, and some tests use credentials, but nothing appears to test this specific feature. I checked in `SubversionPluginTest` in `acceptance-test-harness` too; it has just two extremely simple tests of authenticated access, and no tests of externals.)

oleg-nenashev:
re-:bee:

jglick:
Edits to inline help would probably be appropriate here as well—suggestions from successful users welcome.

jglick:
@reviewbybees done

Awaiting user feedback on effectiveness.

aheritier:
Is it ready to merge cc @jglick ?

jglick:
No, I need to gather feedback on what the issue is.

jglick:
Changing my mind. While there may be more to do, this should be self-contained and potentially helpful.

PR #178, opened 2017-02-28
Using official version of AbstractSampleRepoRule

jglick:
Downstream of https://github.com/jenkinsci/scm-api-plugin/pull/32.

@reviewbybees


jglick:
@reviewbybees done

PR #179, opened 2017-03-15
Fix grammatical error on SVN access exception

jtylr:
Not a huge issue, just something I noticed while using the plugin.


src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -3113,7 +3113,7 @@ public FormValidation checkCredentialsId(StaplerRequest req, @Nonnull Item conte
                     }
                 } catch (SVNException e) {
                     LOGGER.log(Level.SEVERE, e.getErrorMessage().getMessage());
-                    return FormValidation.error("Unable to access to repository");
+                    return FormValidation.error("Unable to access repository");
```
oleg-nenashev: maybe "the repository"

jtylr:
Looks like Jenkins may be stuck with this one.

PR #180, opened 2017-03-20
Update UpdateUpdater.java (Fail build when locked file instead of fresh check out. This is helpful when repositories are very large and takes hours for fresh check out.)

clickjohndahl:
Fail build when locked file instead of fresh check out. This is helpful when repositories are very large and takes hours for fresh check out which flushes configurations as well.

### DIFF
--- a/src/main/java/hudson/scm/subversion/UpdateUpdater.java
+++ b/src/main/java/hudson/scm/subversion/UpdateUpdater.java
@@ -183,7 +183,7 @@ public class UpdateUpdater extends WorkspaceUpdater {
                         
                         //show error instead of check out, useful for big repositories
                          listener.getLogger().println("Workspace appear to be locked, so Failing the build");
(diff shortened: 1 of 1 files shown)


jglick:
Seems like this could break legitimate use cases? Unclear.

oleg-nenashev (changes requested):
Yes, breaking change. Needs a configurable behavior at least

ronaldoooo:
@oleg-nenashev  @jglick I really need this feature to be in the release version. Everytime my workspace get locked by some circumstance, jenkins automaticlly clean my workspace without any suggestions. My whole workspace get lost (Bigger than 200GB) and need a fresh checkout. That kills me.

oleg-nenashev:
@ronaldoooo Hi. Would you be interested to submit a pull request which addresses the review comments? I might be able to merge and release it then

Eccenux:
Stale; possibly not a problem any more.

PR #181, opened 2017-03-27
NoopUpdater

moisei:
https://issues.jenkins-ci.org/browse/JENKINS-2717
built java svn client is very slow for big working copies. 
this change could help to work around this problem by preventing the working copy being updated by the java client. Instead, working copy can be update by the native snv client from pre-scm-build step.

I am attaching a screenshot of sample configuration and also it is committed to src\main\resources\hudson\scm\subversion\NoopUpdater\sample_noop_svn_config.png
sample_noop_svn_config


src/main/java/hudson/scm/subversion/NoopUpdater.java
```diff
@@ -0,0 +1,73 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2004-2010, Sun Microsystems, Inc., Kohsuke Kawaguchi, Fulvio Cavarretta,
```
oleg-nenashev: IMHO License header should be updated

src/main/java/hudson/scm/subversion/NoopUpdater.java
```diff
@@ -0,0 +1,73 @@
...
+
+        @Override
+        public List<External> perform() throws IOException, InterruptedException {
+            return new ArrayList<External>();
+        }
+
+
+    }
+
+    @Extension(ordinal = 100) // this is the default, so given a higher ordinal
```
jglick: Seems to have caused `acceptance-test-harness` failures. I think you meant this to be *negative*. Checking whether that fixes things.

jglick: Was a blind copy-paste from `UpdateUpdater`.

jglick: And since `N` precedes `U`, it was the first choice in the list, which of course breaks every UI-based test!

PR #182, opened 2017-05-01
[JENKINS-26100] Change buildEnvVars to Run.

abayer:
[JENKINS-26100](https://issues.jenkins-ci.org/browse/JENKINS-26100)

NOTE: Findbugs is disabled due to the fact that there are 33 freakin'
findbugs warnings that cause the whole build to error out, and fixing
them is going to take a non-trivial amount of refactoring. I may well
come back to that in this PR, but it may need to wait for another PR.

cc @reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.27-20170501.155323-2</version> <!-- TODO: Switch to 2.27 once https://github.com/jenkinsci/plugin-pom/pull/59 is released -->
+    <version>2.27</version>
     <relativePath />
   </parent>
 
@@ -62,9 +62,9 @@ THE SOFTWARE.
   </scm>
 
   <properties>
-    <jenkins-core.version>2.58-20170501.171905-3</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
-    <jenkins-war.version>2.58-20170501.171927-3</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
-    <java.level>7</java.level>
+    <jenkins-core.version>2.58-20170502.192524-8</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+    <jenkins-war.version>2.58-20170502.192544-8</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+    <java.level>8</java.level>
     <no-test-jar>false</no-test-jar>
     <workflow.version>1.14.2</workflow.version>
     <findbugs.failOnError>false</findbugs.failOnError>
--- a/pom.xml
+++ b/pom.xml
@@ -62,8 +62,7 @@ THE SOFTWARE.
   </scm>
 
   <properties>
-    <jenkins-core.version>2.58-20170502.192524-8</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
-    <jenkins-war.version>2.58-20170502.192544-8</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+    <jenkins.version>2.60</jenkins.version>
     <java.level>8</java.level>
     <no-test-jar>false</no-test-jar>
     <workflow.version>1.14.2</workflow.version>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.23</version>
+    <version>2.27-20170501.155323-2</version> <!-- TODO: Switch to 2.27 once https://github.com/jenkinsci/plugin-pom/pull/59 is released -->
     <relativePath />
   </parent>
 
@@ -62,12 +62,14 @@ THE SOFTWARE.
   </scm>
 
   <properties>
-    <jenkins.version>1.609.3</jenkins.version>
-    <java.level>6</java.level>
+    <jenkins-core.version>2.58-20170501.171905-3</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+    <jenkins-war.version>2.58-20170501.171927-3</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+    <java.level>7</java.level>
     <no-test-jar>false</no-test-jar>
     <workflow.version>1.14.2</workflow.version>
     <findbugs.failOnError>false</findbugs.failOnError>
-    <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
+    <findbugs.skip>true</findbugs.skip> <!-- TODO: Fix the large number of findbugs warnings -->
+    <scm-api-plugin.version>2.1.0</scm-api-plugin.version>
   </properties>
 
   <repositories>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -62,12 +62,14 @@ THE SOFTWARE.
   </scm>
 
   <properties>
-    <jenkins.version>1.609.3</jenkins.version>
-    <java.level>6</java.level>
+    <jenkins-core.version>2.58-20170501.171905-3</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+    <jenkins-war.version>2.58-20170501.171927-3</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+    <java.level>7</java.level>
```
jglick: 8

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -666,12 +666,18 @@ public boolean isFilterChangelog() {
     }
 
     /**
-     * Sets the <tt>SVN_REVISION_n</tt> and <tt>SVN_URL_n</tt> environment variables during the build.
+     * Needed to make Mockito happy due to not picking up the parent implementation.
```
jglick: That is why you need a different name.

pom.xml
```diff
@@ -62,12 +62,14 @@ THE SOFTWARE.
   </scm>
 
   <properties>
-    <jenkins.version>1.609.3</jenkins.version>
-    <java.level>6</java.level>
+    <jenkins-core.version>2.58-20170502.192524-8</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+    <jenkins-war.version>2.58-20170502.192544-8</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
```
jglick: 2.60

pom.xml
```diff
@@ -62,12 +62,13 @@ THE SOFTWARE.
   </scm>
 
   <properties>
-    <jenkins.version>1.609.3</jenkins.version>
-    <java.level>6</java.level>
+    <jenkins.version>2.60</jenkins.version>
```
oleg-nenashev: Needs feedback from the plugin maintainer. I am not comfortable with pushing master version and dropping support of all LTS versions just to get the improvement landed

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -666,12 +666,18 @@ public boolean isFilterChangelog() {
...
-        super.buildEnvVars(build, env);
+    public void buildEnvVars(AbstractBuild<?,?> build, Map<String,String> env) {
+        buildEnvironment(build, env);
+    }
 
+    /**
+     * Sets the <tt>SVN_REVISION_n</tt> and <tt>SVN_URL_n</tt> environment variables during the build.
+     */
+    @Override
+    public void buildEnvironment(Run<?, ?> build, Map<String, String> env) {
```
oleg-nenashev: Why don't you just create this method without override definition? In such case you won't need to bump the core, right?

pom.xml
```diff
@@ -62,12 +62,13 @@ THE SOFTWARE.
...
+    <java.level>8</java.level>
     <no-test-jar>false</no-test-jar>
     <workflow.version>1.14.2</workflow.version>
     <findbugs.failOnError>false</findbugs.failOnError>
-    <scm-api-plugin.version>2.0.8</scm-api-plugin.version>
+    <findbugs.skip>true</findbugs.skip> <!-- TODO: Fix the large number of findbugs warnings -->
```
oleg-nenashev: 🐛 It it is supposed to hurt. Nobody will fix them otherwise

src/main/java/hudson/scm/SubversionSCM.java
abayer: I honestly don't know - I assumed it wouldn't work right if it wasn't explicitly overriding.

oleg-nenashev: IIRC we were doing it many times to pick new APIs in the core without bumping the dependency. CC @jglick to double check

abayer: Well, one problem is that we'd also have to add a `buildEnvVars(AbstractBuild,Map)` method to the SCMs, since they wouldn't be able to rely on the updated `SCM.buildEnvVars(AbstractBuild,Map)` method...

abayer: Ok, a quick test with git seems to suggest this will work, so switching around.

abayer:
Switched back to 1.609.3 baseline etc - but note that there are still tons of findbugs errors. I don't know if those were present beforehand or what.

oleg-nenashev:
> I don't know if those were present beforehand or what.

They were, no worries

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -665,13 +665,15 @@ public boolean isFilterChangelog() {
...
+    }
+
     /**
      * Sets the <tt>SVN_REVISION_n</tt> and <tt>SVN_URL_n</tt> environment variables during the build.
      */
-    @Override
-    public void buildEnvVars(AbstractBuild<?, ?> build, Map<String, String> env) {
-        super.buildEnvVars(build, env);
-
+    public void buildEnvironment(Run<?, ?> build, Map<String, String> env) {
```
oleg-nenashev: I would add info about target Jenkins core version to Javadoc. maybe with "TODO: override when the baseline is above N"

src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
```diff
@@ -231,7 +231,7 @@ protected void retrieve(@CheckForNull SCMSourceCriteria criteria,
                             @NonNull final SCMHeadObserver observer,
                             @CheckForNull SCMHeadEvent<?> event,
                             @NonNull TaskListener listener)
-            throws IOException {
+            throws IOException, InterruptedException {
```
oleg-nenashev: IMHO not required. I also doubt it is binary compatible for methods, which explicitly call SubversionScmSource instead of the method on the parent class.

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -665,13 +665,15 @@ public boolean isFilterChangelog() {
       return filterChangelog;
     }
 
+    @Override
```
jglick:
```java
// TODO 2.60+ delete this overload
```

src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
jglick: If the body does not throw it then it may be removed. It is not a binary compatibility issue since the JVM does not care about checked exceptions.

src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
```diff
@@ -347,7 +347,7 @@ void fetch(@NonNull TaskListener listener,
                @NonNull SortedSet<List<String>> excludedPaths,
                @CheckForNull SCMSourceCriteria branchCriteria,
                @NonNull SCMHeadObserver observer)
-            throws IOException, SVNException {
+            throws IOException, SVNException, InterruptedException {
```
jglick: Ditto.

src/test/java/hudson/scm/SubversionSCMUnitTest.java
```diff
@@ -108,6 +109,7 @@ public void shouldSetEnvironmentVariablesWithMultipleSvnModules() throws IOExcep
     private SubversionSCM mockSCMForBuildEnvVars() {
         SubversionSCM scm = mock(SubversionSCM.class);
         doCallRealMethod().when(scm).buildEnvVars(any(AbstractBuild.class), anyMapOf(String.class, String.class));
+        doCallRealMethod().when(scm).buildEnvironment(any(Run.class), anyMapOf(String.class, String.class));
```
jglick: So does this allow

    mvn -Djenkins.version=2.60 clean test

to work? What about PCT’s

    mvn clean test-compile && mvn -Djenkins.version=2.60 surefire:test

?

src/main/java/hudson/scm/SubversionSCM.java
jglick:
```java
// TODO 2.60+ add @Override
```

src/test/java/hudson/scm/SubversionSCMUnitTest.java
abayer: Yup, works.

src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
abayer: Done.

abayer:
@reviewbybees done

jglick:
Do you need a release cut of this?

oleg-nenashev:
The plugin has no maintainer anymore, so yeah, whomever takes it first

abayer:
@jglick Yup, a release would be handy. =)

James-Dengel:
@jglick Yes that would be great as I'm looking at this right now.

oleg-nenashev:
@James-Dengel It has been already released in Subversion 2.8 from what I see

PR #184, opened 2017-05-29
JENKINS-42186 Fix some potential file leaks

i386:
[JENKINS-42186](https://issues.jenkins-ci.org/browse/JENKINS-42186)

A quick scan of the source code shows that we don't properly clean up instances of `SVNRepository` like we should.

What seems to be happening is that we are creating `SVNRepositoryView` over and over in `retrieve` and not closing it. When this occurs, we end up with many instances of `SVNRepositoryView` open which has file descriptors open to a on disk cache (created by `DBMaker.newFileDB`). This is what eats all the file handles and causes havoc. 

PTAL @recena @jglick 

@reviewbybees

### DIFF
--- a/src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
+++ b/src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
@@ -171,6 +171,7 @@ public class ListSubversionTagsParameterDefinition extends ParameterDefinition {
   @Nonnull public List<String> getTags(@Nullable Job context) {
     List<String> dirs = new ArrayList<String>();
 
+    SVNRepository repo = null;
     try {
       ISVNAuthenticationProvider authProvider = CredentialsSVNAuthenticationProviderImpl.createAuthenticationProvider(
               context, getTagsDir(), getCredentialsId(), null
@@ -178,7 +179,7 @@ public class ListSubversionTagsParameterDefinition extends ParameterDefinition {
       ISVNAuthenticationManager authManager = SubversionSCM.createSvnAuthenticationManager(authProvider);
       SVNURL repoURL = SVNURL.parseURIDecoded(getTagsDir());
 
-      SVNRepository repo = SVNRepositoryFactory.create(repoURL);
+      repo = SVNRepositoryFactory.create(repoURL);
       repo.setAuthenticationManager(authManager);
       SVNLogClient logClient = new SVNLogClient(authManager, null);
       
@@ -194,6 +195,10 @@ public class ListSubversionTagsParameterDefinition extends ParameterDefinition {
       // logs are not translated (IMO, this is a bad idea to translate logs)
       LOGGER.log(Level.SEVERE, "An SVN exception occurred while listing the directory entries at " + getTagsDir(), e);
       return Collections.singletonList("!" + ResourceBundleHolder.get(ListSubversionTagsParameterDefinition.class).format("SVNException"));
+    } finally {
+       if (repo != null) {
+         repo.closeSession();
+       }
     }
 
     // SVNKit's doList() method returns also the parent dir, so we need to remove it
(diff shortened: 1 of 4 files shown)


src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
```diff
@@ -310,6 +311,8 @@ protected SCMRevision retrieve(String unparsedRevision, TaskListener listener) t
             return new SCMRevisionImpl(new SCMHead(base), revision == -1 ? resolvedRevision : revision);
         } catch (SVNException e) {
             throw new IOException(e);
+        } finally {
+            closeSession(repository);
```
i386: I am pretty sure this was the cause of the reported leak on [JENKINS-42186](https://issues.jenkins-ci.org/browse/JENKINS-42186) given that successive calls to `retrieve` seem to be able to reproduce the leak.

src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
```diff
@@ -194,6 +195,10 @@ public DescriptorImpl getDescriptor() {
       // logs are not translated (IMO, this is a bad idea to translate logs)
       LOGGER.log(Level.SEVERE, "An SVN exception occurred while listing the directory entries at " + getTagsDir(), e);
       return Collections.singletonList("!" + ResourceBundleHolder.get(ListSubversionTagsParameterDefinition.class).format("SVNException"));
+    } finally {
+       if (repo != null) {
+         repo.closeSession();
+       }
```
i386: Not needed to fix JENKINS-42186 but is leaking HTTP connections.

src/main/java/jenkins/scm/impl/subversion/SVNRepositoryView.java
```diff
@@ -66,29 +66,29 @@
...
-                if (r <= 0) {
-                    r = (int) TimeUnit2.MINUTES.toMillis(1);
-                }
-                return r;
-            }
-        };
-        repository.setTunnelProvider(SVNWCUtil.createDefaultOptions(true));
-        repository.setAuthenticationManager(sam);
         try {
+            File configDir = SVNWCUtil.getDefaultConfigurationDirectory();
```
i386: Moving everything within the try as a Just In Case.

src/test/java/hudson/scm/SubversionSCMTest.java
```diff
@@ -1223,8 +1223,12 @@ public void invalidTest3936()  throws Exception {
...
     private void attemptAccess(ISVNAuthenticationManager m) throws SVNException {
         SVNRepository repository = SVNRepositoryFactory.create(repo);
-        repository.setAuthenticationManager(m);
-        repository.testConnection();
+        try {
+            repository.setAuthenticationManager(m);
+            repository.testConnection();
+        } finally {
+            repository.closeSession();
+        }
```
i386: Not needed to fix JENKINS-42186 but is leaking HTTP connections during test runs.

src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
```diff
@@ -171,14 +171,15 @@ public DescriptorImpl getDescriptor() {
   @Nonnull public List<String> getTags(@Nullable Job context) {
     List<String> dirs = new ArrayList<String>();
 
+    SVNRepository repo = null;
```
jglick: Better style (IMO) to just have a nested `try`-`finally` block with `repo` as an effectively-`final` variable. Would migrate nicely to `-source 7`, assuming `SVNRepository` is `AutoCloseable`.

src/main/java/jenkins/scm/impl/subversion/SVNRepositoryView.java
jglick: [diff for reviewers](https://github.com/jenkinsci/subversion-plugin/pull/184/files?w=1)

i386:
@recena are you the maintainer of this plugin? If so, what do you think of this change and can we get a release?

recena:
@i386 I was working for some time on this plugin. But clearly, I'm not working now.

recena:
@i386 I could release this plugin when your PR were reviewed and you are confident with the changes.

i386:
Thanks @jglick 👍

PR #185, opened 2017-05-29
JENKINS-25825 Subversion plugin refuses to checkout branch at specifi…

i386:
…c revision when it no longer exists in HEAD

User suggested patch for [JENKINS-25825](https://issues.jenkins-ci.org/browse/JENKINS-25825). Not sure if this is right nor not but thought I could queue it up here for others to review.

@reviewbybees


stephenc (approved):
Seems ok

jglick (commented):
This sounds like something which could, and should, be covered by a new test case.

i386:
FYI this isn't mergable. Another problem reported on the ticket.

oleg-nenashev (changes requested):
NO-MERGE according to the comment from @i386

PR #186, opened 2017-05-29
JENKINS-32302 dead code before Credentials API causes NPE

i386:
It looks like there was some sort of credentials management within this plugin before the credentials api was introduced. Some of that code is still on hot paths and caused the NPE reported in [JENKINS-32302](https://issues.jenkins-ci.org/browse/JENKINS-32302). @stephenc this is an oldie of yours I think. I wonder if its safe just to remove all that migration code and the old credentials impl?

@reviewbybees


i386:
@oleg-nenashev this doesn't drop the code - it just removes updating this map (which is marked as transient) and the migration has already run. So theres no need to update the map anymore.

stephenc:
@oleg-nenashev this code is safe to remove, hence my approval, the other legacy code is not (which is what I was replying to in the text, though I concede that I could have contextualised better)

oleg-nenashev:
Dismissed my review.
@reviewbybees done

i386:
Thanks again @jglick

PR #189, opened 2017-06-20
[JENKINS-35227] added support for AdditionalCredentials to SubversionSCMSource

r-funke:
This fixes issues with multibranch pipeline jobs and SVN externals. The main problem was that SubversionSCMSource did not allow to configure additional credentials, which caused "svn: E200015: ISVNAuthentication" errors. This patch fixes that. The second problem was that during "Scan Multibranch Pipeline" it was just checked whether the revision number of the SVN root directory changed. Thus it did not detect changes in externals. That was fixed by declaring revision number non-deterministic in SCMRevisionImpl, if additional credentials are configured. Then, Jenkins polls for changes in the whole repository.

This should fix JENKINS-35227 and JENKINS-32167. I am not able to post a comment in the bug reports, because I failed to create an account on jenkins.io.


src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -3238,98 +3254,5 @@ private static StandardCredentials lookupCredentials(Item context, String creden
                         CredentialsMatchers.withId(credentialsId));
     }
 
-    public static class AdditionalCredentials extends AbstractDescribableImpl<AdditionalCredentials> {
```
jglick: No, you must leave this class where it was for reasons of settings compatibility. Nothing is stopping `SubversionSCMSource` from reusing it here.

src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
```diff
@@ -136,13 +139,23 @@ public SubversionSCMSource(String id, String remoteBase, String credentialsId, S
...
+    public SubversionSCMSource(String id, String remoteBase) {
+        super(id);
+        this.remoteBase = StringUtils.removeEnd(remoteBase, "/") + "/";
+        this.additionalCredentials =  new ArrayList<AdditionalCredentials>();
     }
 
     @DataBoundConstructor
     @SuppressWarnings("unused") // by stapler
-    public SubversionSCMSource(String id, String remoteBase) {
+    public SubversionSCMSource(String id, String remoteBase, List<AdditionalCredentials> additionalCredentials) {
```
jglick: No, this should be in a `@DataBoundSetter` only.

src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
```diff
@@ -701,16 +723,29 @@ public SubversionSCM build(@NonNull SCMHead head,
...
         }
 
         public long getRevision() {
             return revision;
         }
 
+        public boolean isDeterministic() {
+            //non-deterministic when using externals, to force Jenkins to recurse into subdirectories
+            if(usesExternals) {
+                return false;
```
jglick: Nondeterministic revisions are acceptable only for SCMs which inherently cannot support determinism, like CVS I suppose (though even there `-D` timestamps may offer a solution). In the case of Subversion with externals, we would rather need to extend `SCMRevisionImpl` to encode the revision number in each external encountered.

src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
```diff
@@ -273,7 +294,7 @@ protected SCMRevision retrieve(@NonNull SCMHead head, @NonNull TaskListener list
             String repoPath = SubversionSCM.DescriptorImpl.getRelativePath(repoURL, repository.getRepository());
             String path = SVNPathUtil.append(repoPath, head.getName());
             SVNRepositoryView.NodeEntry svnEntry = repository.getNode(path, -1);
-            return new SCMRevisionImpl(head, svnEntry.getRevision());
+            return new SCMRevisionImpl(head, svnEntry.getRevision(), (additionalCredentials.size()>0));
```
jglick: This is not precise. Additional credentials might have been specified but not actually used. More plausibly, the server may use externals which simply do not need any distinct credentials.

jglick (dismissed):

r-funke:
@jglick thank you for your input. I've moved AdditionalCredentials back in place.

Regarding the non-deterministic revision, it was an easy solution (or workaround) for Jenkins not detecting changes in externals. Of course, the clean solution would be to extend SCMRevisionImpl. But to me it seems that it requires quite a lot of effort to get that done. 

Would it be okay to have a checkbox in the config and let the user specify whether isDeterministic() should return true or false?

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -297,7 +297,7 @@ public SubversionSCM(List<ModuleLocation> locations,
      */
     public SubversionSCM(List<ModuleLocation> locations,
             boolean useUpdate, SubversionRepositoryBrowser browser, String excludedRegions, String excludedUsers, String excludedRevprop, String excludedCommitMessages) {
-    	this(locations, useUpdate, false, browser, excludedRegions, excludedUsers, excludedRevprop, excludedCommitMessages);
+        this(locations, useUpdate, false, browser, excludedRegions, excludedUsers, excludedRevprop, excludedCommitMessages);
```
oleg-nenashev: May cause merge conflicts without a reason

src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
```diff
@@ -121,6 +122,8 @@
     private final String remoteBase;
 
     private String credentialsId = ""; // TODO null would be a better default, but need to check null safety on usages
+    
+    private List<AdditionalCredentials> additionalCredentials;
```
oleg-nenashev: from what I see it still may be null when you read data from the disk

src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
```diff
@@ -159,6 +165,14 @@ public String getCredentialsId() {
     public void setCredentialsId(String credentialsId) {
         this.credentialsId = credentialsId;
     }
+    @DataBoundSetter
+    public void setAdditionalCredentials(List<AdditionalCredentials> additionalCredentials) {
+        this.additionalCredentials = additionalCredentials;
+    }
+    //without getter jelly will crash
+    public List<AdditionalCredentials> getAdditionalCredentials() {
+        return(additionalCredentials);
```
oleg-nenashev: FindBugs will grumble about exposing external implementation. Either make it `@Restricted` for Jelly only or wrap with `unmodifyableList`

src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
```diff
@@ -701,16 +716,29 @@ public SubversionSCM build(@NonNull SCMHead head,
          * The subversion revision.
          */
         private long revision;
+        /**
+         * Indicates whether externals are used. If so, revision is considered non-deterministic.
+         */
+        private boolean usesExternals;
 
-        public SCMRevisionImpl(SCMHead head, long revision) {
+        public SCMRevisionImpl(SCMHead head, long revision, boolean usesExternals) {
```
oleg-nenashev: Breaking change, the class was public

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -403,6 +410,14 @@ public SubversionSCM(String[] svnUrls, String[] credentialIds, String[] locals)
         this(ModuleLocation.parse(svnUrls, credentialIds, locals, null,null), true, false, null, null, null, null, null);
     }
 
+
+    /**
+     * Convenience constructor, especially during testing.
+     */
+    public SubversionSCM(String[] svnUrls, String[] credentialIds, String[] locals, List<AdditionalCredentials> additionalCredentials) {
```
oleg-nenashev: I do not believe new methods are really required if you implement proper null handling in the original constructors. You say these methods are useful for testing, but there is no autotests

src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
```diff
@@ -136,15 +139,18 @@ public SubversionSCMSource(String id, String remoteBase, String credentialsId, S
         setCredentialsId(credentialsId);
         setIncludes(StringUtils.defaultIfEmpty(includes, DescriptorImpl.DEFAULT_INCLUDES));
         setExcludes(StringUtils.defaultIfEmpty(excludes, DescriptorImpl.DEFAULT_EXCLUDES));
+        this.additionalCredentials =  new ArrayList<AdditionalCredentials>();
```
oleg-nenashev: Usage of this collection is ineffective if you do not plan to modify it on the flight. Use `Collections.emptyList()` instead in the code

oleg-nenashev (changes requested):
This PR introduces a breaking change + additionalCredentials handling seems to be error-prone and redundant in some cases.

I would also expect to see a test suite for the change

src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
r-funke: Thank you for your comments. How do I make it restricted? Does adding "@Restricted(NoExternalUse.class)" work?

r-funke:
Would it help to remove the changes to SCMRevisionImpl and isDeterministic() from this PR to get it accepted? Then this PR would only cover AdditionalCredentials for SubversionSCMSource, i.e. for multibranch pipeline jobs, which would solve JENKINS-35227.

src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
oleg-nenashev: yes

oleg-nenashev:
> So maybe I should create an issue for that first.

+1. An issue would be useful. The scope of this change is not that big, so probably we could have two fixes in a single PR

oleg-nenashev:
Separate pull requests would be better though

r-funke:
Okay, this PR is now only about additional credentials. Is it good enough to be accepted?

src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
```diff
@@ -159,6 +165,18 @@ public String getCredentialsId() {
...
+    public void setAdditionalCredentials(List<AdditionalCredentials> additionalCredentials) {
+        this.additionalCredentials = additionalCredentials;
+    }
+    //without getter jelly will crash
+    @Restricted(NoExternalUse.class)
+    public List<AdditionalCredentials> getAdditionalCredentials() {
+        if(additionalCredentials==null) {
+            return Collections.emptyList();
+        }
+        return(additionalCredentials);
```
oleg-nenashev: would be better to just `return Collections.unmodifyableList(additionalCredentials)`

oleg-nenashev:
Yes, it looks much better now. I probably could live even without tests for it. @jglick WDYT?

r-funke:
Any news on this issue?

DavidA2014:
Any news on this issue?  This fix would help me.

jglick:
Still would benefit from tests. Probably ought to be migrated to traits in a newer `scm-api`. Requires a maintainer to do this work I think.

imakowski:
When this will be released?

larsbier:
I try to understand, what's happened and why :-) I have a provider, which needs username/password and PIN-protected client certificate. And I have to move my multibranch pipeline to this svn provider.
So I cannot say, that this request is obsolete :-)

Eccenux:
Stale; possibly not a problem any more.

PR #190, opened 2017-06-25
Repository session was not closed for unexpected exceptions

trejkaz:
Previous code caught two specific checked exceptions and closed the repository session only for those. But RuntimeException could occur as well, so the new pattern will work in all cases where success did not occur.


trejkaz:
A bit of background here is that we're seeing Jenkins keeping SVN repositories locked, but in a non-reproducible way which prevents us opening an issue for it. I found this in a quick audit of the code for improper resource handling. It may not be the issue, but it's at least _an_ issue. I should probably check SVNKit itself too.

oleg-nenashev:
Retriggered CI

PR #191, opened 2017-07-11
[JENKINS-45362] Workaround for not detecting changes in externals

r-funke:
Like discussed in PR #189 the clean solution would be to extend SCMRevisionImpl to encode the revisions of all externals. But setting isDeterministic() to false is a simple workaround that solves the issue.

### DIFF
--- a/src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
+++ b/src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
@@ -68,6 +68,8 @@ import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.tmatesoft.svn.core.SVNDirEntry;
 import org.tmatesoft.svn.core.SVNException;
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
@jenkinsci/code-reviewers

daniel-beck:
I'm not sure it's a great idea to add a workaround with a UI option.

r-funke:
If #189 gets accepted, what about enabling the workaround whenever additional credentials have been enabled? @jglick commented on this that 1) additional credentials may have been specified but not used, and 2) there may be externals which do not require additional credentials. Regarding 1) I think that this will rarely happen. But if it does, the workaround doesn't seem to break anything. Regarding 2) my experience is that it is always necessary to specify additional credentials when using externals to avoid ISVNAuthentication exception (JENKINS-35227).

r-funke:
Any further suggestions?

imakowski:
Any news about this?

theKBro:
this is still an issue.... 
for each external even if it is from the same svn server we get an error
```ERROR: Failed to check repository revision for https://subversion/f_e/Standard_Development/EmbeddedSoftwareArchitecture/trunk/UserArea/CustomerEdsBlackList.txt
org.tmatesoft.svn.core.SVNCancelException: svn: E200015: E200015: ISVNAuthentication provider did not provide credentials; HTTP authorization cancelled.
svn: E200015: ISVNAuthentication provider did not provide credentials; HTTP authorization cancelled.```
the workaround for this is adding the normal credentials again as additional credentials...
Are there plans to finally fix this?

PR #194, opened 2017-08-14
Added a Jenkinsfile

jglick:
@reviewbybees

PR #193, opened 2017-08-08
POM update

jglick:
@reviewbybees

PR #195, opened 2017-08-14
Skip tests on CI when Subversion is not installed locally

jglick:
Amends #194 which was merged prematurely. Subsumes #193. Quick fix to get builds passing. Better would be to improve tests to run portably, as mentioned in comments.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -121,16 +121,16 @@ THE SOFTWARE.
     </dependency>
 
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <version>4.11</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
       <version>1.9.5</version>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.hamcrest</groupId>
+          <artifactId>hamcrest-core</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.powermock</groupId>
--- a/src/test/java/hudson/scm/AbstractSubversionTest.java
+++ b/src/test/java/hudson/scm/AbstractSubversionTest.java
@@ -1,40 +1,47 @@
 package hudson.scm;
 
 import hudson.ClassicPluginStrategy;
+import hudson.FilePath;
 import hudson.Launcher.LocalLauncher;
 import hudson.Proc;
-import hudson.scm.SubversionSCM.DescriptorImpl;
 import hudson.util.StreamTaskListener;
 
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 import net.sf.json.JSONObject;
 
-import org.jvnet.hudson.test.HudsonHomeLoader.CopyExisting;
-import org.jvnet.hudson.test.HudsonTestCase;
 import org.kohsuke.stapler.StaplerRequest;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.ServerSocket;
 import java.net.Socket;
 import java.net.URL;
 import org.junit.AssumptionViolatedException;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.rules.TemporaryFolder;
+import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.JenkinsRule;
 
 /**
  * Base class for Subversion related tests.
  *
  * @author Kohsuke Kawaguchi
  */
-public abstract class AbstractSubversionTest extends HudsonTestCase  {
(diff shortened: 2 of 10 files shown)


src/test/java/hudson/scm/AbstractSubversionTest.java
```diff
@@ -57,12 +67,9 @@ public static Proc runSvnServe(URL zip) throws Exception {
      * Runs svnserve to serve the specified directory as a subversion repository.
      */
     public static Proc runSvnServe(File repo) throws Exception {
```
jglick: Probably this utility could be better moved to `SubversionSampleRepoRule` as well, but leaving for another day.

jglick:
[Latest test failures](https://ci.jenkins.io/job/Plugins/job/subversion-plugin/job/PR-195/4/testReport/) seem to be random

```
java.net.SocketException: Permission denied: listen
```

pom.xml
```diff
@@ -121,16 +121,20 @@ THE SOFTWARE.
...
-      <version>4.11</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
       <version>1.9.5</version>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
```
oleg-nenashev: It makes sense to mention conflicted versions in comments when you fix upper bounds

pom.xml
```diff
@@ -143,6 +147,12 @@ THE SOFTWARE.
       <artifactId>powermock-api-mockito</artifactId>
       <version>1.5</version>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.mockito</groupId>
+          <artifactId>mockito-all</artifactId>
```
oleg-nenashev: Ideally this dependency should be removed at all from the plugin. Mockito Core should be enough
https://github.com/mockito/mockito/wiki/Declaring-mockito-dependency

src/test/java/hudson/scm/AbstractSubversionTest.java
```diff
@@ -46,23 +53,41 @@ protected void configureSvnWorkspaceFormat(int format) throws Exception {
...
-    	this.descriptor.configure(req, formData);
+    	r.jenkins.getDescriptorByType(SubversionSCM.DescriptorImpl.class).configure(req, formData);
+    }
+
+    public static void checkForSvnServe() throws InterruptedException {
+        LocalLauncher launcher = new LocalLauncher(StreamTaskListener.fromStdout());
+        try {
+            launcher.launch().cmds("svnserve","--help").start().join();
+        } catch (IOException e) {
+            // TODO better to add a docker-fixtures test dep so CI builds can run these tests
```
oleg-nenashev: +1, fine for now

src/test/java/hudson/scm/SVNWorkingCopyTest.java
```diff
@@ -23,57 +23,56 @@
...
-import org.tmatesoft.svn.core.wc.SVNStatusClient;
-import org.tmatesoft.svn.core.wc2.SvnGetStatus;
-
-import java.io.File;
-import org.junit.Ignore;
 
 /**
  * @author schristou88
  */
 @Ignore("TODO org.tmatesoft.svn.core.SVNException: svn: E175002: PROPFIND of '/trunk/hudson/test-projects/trivial-ant': 405 Method Not Allowed (https://svn.jenkins-ci.org)")
```
oleg-nenashev: Do we still need to ignore it? CC @christ66

jglick:
> Maybe the flaky test should be just disabled

It is not a specific test. On that build, several tests failed, including `InjectedTest`, and not because of test code.

pom.xml
jglick: The `objenesis` conflict is with PowerMock. This is pretty standard—every plugin using PowerMock winds up with the same conflict. Ditto `hamcrest-core`, conflicting with a newer one from `junit` / `jenkins-test-harness`.

pom.xml
jglick: It is possible a newer version of `powermock-api-mockito` fixes their dependencies; I did not want to upgrade PowerMock here.

Anyone publishing `*-all` artifacts on Central with `jar` packaging should be given a stern lecture…

src/test/java/hudson/scm/SVNWorkingCopyTest.java
jglick: We do: `svn.jenkins-ci.org` is long dead. Another case where we need to use `docker-fixtures` to set up a test server.

(There is another test floating around here that connects to `subversion.tigris.org`, which still works, but is poor style.)

pom.xml
jglick: To be clear, adding this `exclusion` _does_ remove this dependency from the plugin.

oleg-nenashev: OK, let's keep it as is for now

oleg-nenashev: Race condition in comments

PR #136, opened 2015-10-14
[JENKINS-14541] --quiet mode for svn checkout / update

zionyx:
Adds an option in Subversion scm config to reduce significant amount of logs produced during build when checking out / updating working copy.

Includes both English and traditional Chinese help files / descriptions.
Also added a unit test for this feature.

https://issues.jenkins-ci.org/browse/JENKINS-14541

PR #198, opened 2017-10-19
[JENKINS-14541] --quiet mode for svn checkout / update

zionyx:
Adds an option in Subversion scm config to reduce significant amount of logs produced during build when checking out / updating working copy.

Includes both English and traditional Chinese help files / descriptions.
Also added a unit test for this feature.

https://issues.jenkins-ci.org/browse/JENKINS-14541

CC @recena, @jglick, @oleg-nenashev, @TomaszHal 

See https://github.com/jenkinsci/subversion-plugin/pull/136

### DIFF
--- a/src/main/java/hudson/scm/SubversionEventHandlerImpl.java
+++ b/src/main/java/hudson/scm/SubversionEventHandlerImpl.java
@@ -36,9 +36,12 @@ public class SubversionEventHandlerImpl extends SVNEventAdapter {
 
     protected final File baseDir;
 
-    public SubversionEventHandlerImpl(PrintStream out, File baseDir) {
+    protected final boolean quietOperation;
+
+    public SubversionEventHandlerImpl(PrintStream out, File baseDir, boolean quietOperation) {
         this.out = out;
         this.baseDir = baseDir;
+        this.quietOperation = quietOperation;
     }
 
     public void handleEvent(SVNEvent event, double progress) throws SVNException {
@@ -54,6 +57,10 @@ public class SubversionEventHandlerImpl extends SVNEventAdapter {
         }
 
         SVNEventAction action = event.getAction();
+        if (quietOperation && (action != SVNEventAction.UPDATE_COMPLETED)) {
+            //  Skips logging
+            return;
+        }
 
         {// commit notifications
             if (action == SVNEventAction.COMMIT_ADDED) {
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -245,6 +245,7 @@ public class SubversionSCM extends SCM implements Serializable {
 
     private boolean ignoreDirPropChanges;
     private boolean filterChangelog;
+    private boolean quietOperation;
 
     /**
      * A cache of the svn:externals (keyed by project).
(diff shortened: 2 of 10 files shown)


src/main/java/hudson/scm/SubversionEventHandlerImpl.java
```diff
@@ -36,9 +36,12 @@
 
     protected final File baseDir;
 
-    public SubversionEventHandlerImpl(PrintStream out, File baseDir) {
+    protected final boolean quietOperation;
```
oleg-nenashev: I would rather suggest using private fields and the getter. Otherwise the field cannot be removed in a compatible way if the configuration structure changes

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -953,13 +979,15 @@ public boolean requiresWorkspaceForPolling() {
     private static class CheckOutTask extends UpdateTask implements FileCallable<List<External>> {
         private final UpdateTask task;
 
-         public CheckOutTask(Run<?, ?> build, SubversionSCM parent, ModuleLocation location, Date timestamp, TaskListener listener, EnvVars env) {
+        public CheckOutTask(Run<?, ?> build, SubversionSCM parent, ModuleLocation location, Date timestamp,
```
oleg-nenashev: class is private, so it's fine

src/main/resources/hudson/scm/SubversionSCM/help-quietOperation.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    Mimics subversion command line "--quiet" parameter for <b>check-out / update</b> operations to help keep the output shorter.<br>
+    <i>Prints nothing, or only summary information.</i>
+</div>
```
oleg-nenashev: nice2have: line break

recena:
```
<div>
    <p>Mimics subversion command line <code>--quiet</code> parameter for <strong>check-out / update</strong> operations to help keep the output shorter. Prints nothing, or only summary information.</p>
</div>
```

src/main/java/hudson/scm/SubversionEventHandlerImpl.java
zionyx: Alright. It will be updated in the following commit.

src/main/java/hudson/scm/SubversionSCM.java
zionyx: Hmm... So, ignore the previous suggestion, private + getter?

zionyx:
Forced a rebuild of the failing build and all looked good. Please merge when possible. Thank you!

jglick:
Unfortunately this regressed `CredentialsExternalsTest` so it should be backed out unless the fix is obvious.

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -338,12 +339,24 @@ public SubversionSCM(List<ModuleLocation> locations, WorkspaceUpdater workspaceU
...
+    /**
+     *  @deprecated by quietOperation
+     */
+    public SubversionSCM(List<ModuleLocation> locations, WorkspaceUpdater workspaceUpdater,
+            SubversionRepositoryBrowser browser, String excludedRegions, String excludedUsers,
+            String excludedRevprop, String excludedCommitMessages,
+            String includedRegions, boolean ignoreDirPropChanges, boolean filterChangelog,
+            List<AdditionalCredentials> additionalCredentials) {
+        this(locations, workspaceUpdater, browser, excludedRegions, excludedUsers, excludedRevprop, excludedCommitMessages, 
+            includedRegions, ignoreDirPropChanges, false, null, false);
```
jglick: Presumably this should have been

```java
…, filterChangelog, additionalCredentials, false);
```

Will check that.

src/main/java/hudson/scm/SubversionEventHandlerImpl.java
```diff
@@ -36,9 +36,12 @@
 
     protected final File baseDir;
 
-    public SubversionEventHandlerImpl(PrintStream out, File baseDir) {
```
jglick: #203

src/main/java/hudson/scm/subversion/SubversionUpdateEventHandler.java
```diff
@@ -63,8 +63,9 @@
      */
     private final String modulePath;
     
-    public SubversionUpdateEventHandler(PrintStream out, List<External> externals, File moduleDir, String modulePath) {
-        super(out,moduleDir);
```
jglick: #203

PR #199, opened 2017-11-03
[JENKINS-24802] notifyCommit don't trigger a build

kenji21:
Moved try/catch inside the loop to have next ModuleLocation checked if
a SVNCancelExcpetion is raised
Updated error message helping to diagnose which job(s) causes
SVNCancelException(s)


oleg-nenashev:
Formatting-free diff: https://github.com/jenkinsci/subversion-plugin/pull/199/files?w=1

PR #200, opened 2017-11-03
Make as-it-is checkout only files initally

mbirtwell:
SVNKit checks out to infinity by default if unknown is given as the
depth to checkout. If you're using as-it-is you probably don't want the
whole source tree, so change as-it-is to set the depth to files for
checkout and unknown for updates.

Not sure how flexible to be here some people might really want immediates or empty instead of files. But files is probably close enough isn't it. There could be an extra setting for this but I don't think that necessary.

### DIFF
--- a/src/main/java/hudson/scm/subversion/WorkspaceUpdater.java
+++ b/src/main/java/hudson/scm/subversion/WorkspaceUpdater.java
@@ -200,6 +200,7 @@ public abstract class WorkspaceUpdater extends AbstractDescribableImpl<Workspace
          *      module location depth options or SVNDepth.fromString directly if you've taken the
          *      depth from another source.
          */
+        @Deprecated
         protected static SVNDepth getSvnDepth(String name) {
             return SVNDepth.fromString(name);
         }
--- a/src/main/java/hudson/scm/subversion/WorkspaceUpdater.java
+++ b/src/main/java/hudson/scm/subversion/WorkspaceUpdater.java
@@ -190,15 +190,6 @@ public abstract class WorkspaceUpdater extends AbstractDescribableImpl<Workspace
             return r;
         }
 
-        /**
-         * Returns {@link org.tmatesoft.svn.core.SVNDepth} by string value.
-         *
-         * @return {@link org.tmatesoft.svn.core.SVNDepth} value.
-         */
-        protected static SVNDepth getSvnDepth(String name) {
-            return SVNDepth.fromString(name);
-        }
-
         private static final long serialVersionUID = 1L;
     }
 }
--- a/src/main/java/hudson/scm/subversion/WorkspaceUpdater.java
+++ b/src/main/java/hudson/scm/subversion/WorkspaceUpdater.java
@@ -190,6 +190,20 @@ public abstract class WorkspaceUpdater extends AbstractDescribableImpl<Workspace
             return r;
         }
 
+        /**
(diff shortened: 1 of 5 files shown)


mbirtwell:
I've added another similar fix to this pull request  to make the use svn update with revert first setting behave as I would expect in an as-it-is module.

When unknown is passed to the SvnRevert operation of SVNKit it's coerced
to empty, which means nothing actually gets reverted. So instead of
passing unknown pass infinity for revert as the svn depth. This seems to
do what one would want i.e. revert what ever happens to be there but not
checkout any additional files.

src/main/java/hudson/scm/subversion/WorkspaceUpdater.java
```diff
@@ -190,15 +190,6 @@ protected SVNRevision getRevision(ModuleLocation l) {
             return r;
         }
 
-        /**
-         * Returns {@link org.tmatesoft.svn.core.SVNDepth} by string value.
-         *
-         * @return {@link org.tmatesoft.svn.core.SVNDepth} value.
-         */
-        protected static SVNDepth getSvnDepth(String name) {
```
oleg-nenashev: IIUC it breaks the binary compatibility as it confirmed by the changes above.
WorkspaceUpdater implementations may exist outside the plugin, so it needs to be handled carefully

mbirtwell: I don't fully understand what contract we need to provide here. Is it:
1) we just need to continue providing any accessible methods, that people might be using so that we don't break stuff.
2) we need to do 1 + continue to call the functions in a similar way to before so that they can be used as an extension point. 

If 1 then obviously I just need to add the method back. Should I also mark it as deprecated. 
If 2 then I'll also have to change all my getSvnDepthFor* function to getDepthOptionFor* and wrap their invocations in calls the the getSvnDepth function. 

Which do you think is better?

oleg-nenashev: Jenkins project uses the approach number 1, because we cannot guarantee that plugins maintainers deliver the required fixes for N + 1 and then that the users do updates properly. Breaking changes are being delivered only in major releases (first-digit change), but there is no justification of doing such release in the SVN plugin while it has no active maintainer

src/main/java/hudson/scm/subversion/WorkspaceUpdater.java
```diff
@@ -194,6 +194,11 @@ protected SVNRevision getRevision(ModuleLocation l) {
          * Returns {@link org.tmatesoft.svn.core.SVNDepth} by string value.
          *
          * @return {@link org.tmatesoft.svn.core.SVNDepth} value.
+         *
+         * @deprecated as of 2.10
+         *      Use SubversionSCM.ModuleLocation.getSVNDepthFor* functions to correctly interpret
+         *      module location depth options or SVNDepth.fromString directly if you've taken the
+         *      depth from another source.
          */
         protected static SVNDepth getSvnDepth(String name) {
```
oleg-nenashev: I would recommend to also add a `@Deprecated` annotation so that Jenkins tooling will be able to find deprecated usages

mbirtwell:
Done, my IDE treated the method as if it had been annotated as soon as I added the comment and I didn't know that wasn't standard.

mbirtwell:
Thanks both of you

PR #201, opened 2017-12-11
Updated baseline and parent and solving some FindBugs warnings

jglick:
Trying to actually solve the bug worked around in core prior to https://github.com/jenkinsci/jenkins/pull/3192.

@reviewbybees @jenkinsci/code-reviewers


src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -2580,7 +2573,7 @@ private static boolean isAuthenticationFailedError(SVNCancelException e) {
 
     @CheckForNull
     private static DescriptorImpl descriptor() {
-        Jenkins instance = Jenkins.getInstance();
+        Jenkins instance = Jenkins.getInstanceOrNull();
```
jglick: The key fix.

src/main/java/hudson/scm/SubversionWorkspaceSelector.java
```diff
@@ -102,9 +103,9 @@ public Collection getEnabledFactories(File path, Collection factories, boolean w
     public static volatile int workspaceFormat = SVNAdminArea14.WC_FORMAT;
 
     public static void syncWorkspaceFormatFromMaster() {
-        Hudson h = Hudson.getInstance();
-        if (h!=null)
-            workspaceFormat = h.getDescriptorByType(SubversionSCM.DescriptorImpl.class).getWorkspaceFormat();
+        Jenkins j = Jenkins.getInstanceOrNull();
```
jglick: This too, perhaps. Really this code is a legacy design that caused problems from SECURITY-144 and needs to be redesigned.

src/main/java/hudson/scm/listtagsparameter/ListSubversionTagsParameterDefinition.java
```diff
@@ -223,13 +222,7 @@ public DescriptorImpl getDescriptor() {
     }
 
     // SVNKit's doList() method returns also the parent dir, so we need to remove it
-    if(dirs != null) {
-      removeParentDir(dirs);
-    }
-    else {
```
jglick: Apparently dead code path—none of the three definitions above could return null.

src/main/java/hudson/scm/browsers/AbstractSventon.java
```diff
@@ -59,7 +59,7 @@ public AbstractSventon(URL url, String repositoryInstance) throws MalformedURLEx
         // normalize
         repositoryInstance = repositoryInstance.trim();
 
-        this.repositoryInstance = repositoryInstance == null ? "" : repositoryInstance;
```
oleg-nenashev: Yes, it was never going to work - the code was introduced in https://github.com/jenkinsci/subversion-plugin/blame/master/src/main/java/hudson/scm/browsers/AbstractSventon.java

src/main/java/hudson/scm/SubversionWorkspaceSelector.java
stephenc: Are we sure this code path isn’t visited by the agent jvm?

src/main/java/hudson/scm/SubversionSCM.java
stephenc: Are we sure this code path is not reachable by the agent jvm?

jglick: To the contrary, according to [JENKINS-34857](https://issues.jenkins-ci.org/browse/JENKINS-34857), it _is_ reachable from an agent JVM:

```
java.lang.IllegalStateException: Jenkins has not been started, or was already shut down
	at jenkins.model.Jenkins.getInstance(Jenkins.java:766)
	at hudson.scm.SubversionSCM.descriptor(SubversionSCM.java:2544)
	at hudson.scm.SubversionSCM.createDefaultSVNOptions(SubversionSCM.java:1043)
	at hudson.scm.SubversionSCM.createClientManager(SubversionSCM.java:1033)
	at hudson.scm.SubversionSCM$CheckOutTask.invoke(SubversionSCM.java:965)
	at hudson.scm.SubversionSCM$CheckOutTask.invoke(SubversionSCM.java:945)
	at …
	at ......remote call to VM177_Win7_64bits(Native Method)
	at …
	at hudson.FilePath.act(FilePath.java:996)
	at hudson.scm.SubversionSCM.checkout(SubversionSCM.java:894)
	at hudson.scm.SubversionSCM.checkout(SubversionSCM.java:830)
	at org.jenkinsci.plugins.multiplescms.MultiSCM.checkout(MultiSCM.java:143)
	at hudson.scm.SCM.checkout(SCM.java:485)
	at hudson.model.AbstractProject.checkout(AbstractProject.java:1269)
	at …
```

which is exactly why I am switching to `getInstanceOrNull`—using `getInstance` here is illegal.

As noted in the other comment, it would of course be better to rewrite this code to have the `CheckOutTask` receive default SVN options as a constructor argument on the master side and have them be serialized in the normal way. That would be a more intrusive change, though.

src/main/java/hudson/scm/SubversionWorkspaceSelector.java
jglick: Ditto. It _is_ called from the agent side AFAICT, which is why I am using `getInstanceOrNull`.

stephenc: So you are going to try and load Servlet Spec classes on the agent JVM... that's not good.

You cannot invoke any `Jenkins` method safely from an agent

src/main/java/hudson/scm/SubversionSCM.java
jglick: Reproducible for example from `PerJobCredentialStoreTest.remoteBuild` with patched core unless switched to `getInstanceOrNull`.

src/main/java/hudson/scm/SubversionWorkspaceSelector.java
jglick: No, you cannot, which is why this code should be rewritten. But I am leaving that for another day.

jglick: To clarify: `Jenkins.getInstanceOrNull` will normally work OK from an agent (returning null). It _could_ produce Servlet-related linkage errors in relatively obscure environments. But if that was an issue, it has been an issue in this plugin for years. I am just updating to make sure that we do not run into problems with a version of Jenkins core that enforces the documented behavior of `getInstance`.

pom.xml
```diff
@@ -180,19 +180,27 @@ THE SOFTWARE.
...
+      <version>2.0.3</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
     <plugins>
       <plugin>
         <groupId>org.jenkins-ci.tools</groupId>
         <artifactId>maven-hpi-plugin</artifactId>
```
oleg-nenashev: Is it required anymore? Not sure we need to mask classes

src/main/java/hudson/scm/SubversionSCM.java
oleg-nenashev: LGTM

pom.xml
jglick: No idea, but I did not see any reason to touch it in this PR.

src/main/java/hudson/scm/SubversionWorkspaceSelector.java
jglick: Agreed: [JENKINS-48543](https://issues.jenkins-ci.org/browse/JENKINS-48543)

jglick:
@reviewbybees done

pom.xml
oleg-nenashev: Well, you touched it. See the lines below

jglick: I touched some plugin configuration as part of routine POM cleanup. Changing class loading behavior for one of this plugin’s libraries is a much riskier change, and totally unrelated to the rest of this PR.

PR #202, opened 2017-12-12
[JENKINS-14541] Mistaken constructor delegation

jglick:
Corrects a mistake in #198.

@reviewbybees @zionyx


zionyx:
Oops. Thanks for that!

oleg-nenashev:
2.10.1 is out. My bad, the tests were passing locally and the CI was dead

jglick:
This test was skipped on CI because it requires `svnserve` to be installed. It could probably be rewritten to use `docker-fixtures`. Same for any of the other tests that either do not use an agent, or can work with an agent which is actually local (as opposed to itself containerized).

PR #203, opened 2017-12-14
[JENKINS-48546] - Restore the binary compatibility with 2.9

oleg-nenashev:
I missed a binary compatibility issue during the code reviews. We should employ more robots to do that, right? :(

https://issues.jenkins-ci.org/browse/JENKINS-48546

@reviewbybees @jglick

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -979,7 +979,7 @@ public class SubversionSCM extends SCM implements Serializable {
     private static class CheckOutTask extends UpdateTask implements FileCallable<List<External>> {
         private final UpdateTask task;
 
-        public CheckOutTask(Run<?, ?> build, SubversionSCM parent, ModuleLocation location, Date timestamp,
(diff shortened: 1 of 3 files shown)


jglick:
I guess this is another regression from #198? CC @zionyx

oleg-nenashev:
It is

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -979,7 +979,7 @@ public boolean requiresWorkspaceForPolling() {
     private static class CheckOutTask extends UpdateTask implements FileCallable<List<External>> {
         private final UpdateTask task;
 
-        public CheckOutTask(Run<?, ?> build, SubversionSCM parent, ModuleLocation location, Date timestamp,
+        CheckOutTask(Run<?, ?> build, SubversionSCM parent, ModuleLocation location, Date timestamp,
```
jglick: Unrelated I guess.

oleg-nenashev: just wanted to make it explicit, because the method signature has also changed

oleg-nenashev:
@reviewbybees done
🚢 🇮🇹

PR #204, opened 2018-01-18
Add new option "cancelProcessOnExternalsFail" for ModuleLocation

Liyaa:
which determines if the process should be cancelled when checkout/update svn:externals failed.

Sometimes, when svn:externals checkout/update failed, the artifacts generated is likely to fall short of expectations, but now it's ignored by default and no message can found in the log.

### DIFF
--- a/src/main/resources/hudson/scm/SubversionSCM/ModuleLocation/help-cancelProcessOnExternalsFail.html
+++ b/src/main/resources/hudson/scm/SubversionSCM/ModuleLocation/help-cancelProcessOnExternalsFail.html
@@ -1,0 +1,3 @@
+<div>
+    Determines if the process should be cancelled when checkout/update svn:externals failed. Will work when "Ignore externals" box is not checked.
+</div>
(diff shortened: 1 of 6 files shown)


src/main/resources/hudson/scm/SubversionSCM/ModuleLocation/help-cancelProcessOnExternalsFail.html
```diff
@@ -0,0 +1,3 @@
+<div>
+    Determines if the process should be cancelled when checkout/update svn:externals failed. Will work when "Ignore externals" box is not checked.
```
oleg-nenashev: Maybe it needs to be clarified to explicitly say that the default behavior is to fail the checkout

oleg-nenashev:
CC @kuisathaverat

src/main/resources/hudson/scm/SubversionSCM/ModuleLocation/help-cancelProcessOnExternalsFail.html
Liyaa: Sorry for ignoring the code formatting while my IDE using different style, already correct it, and default behavior is clarified too.

Liyaa:
@oleg-nenashev @kuisathaverat Thank you for your approval! Is there any other work I need to supplement? Can these commits be merged for now? ^_^

oleg-nenashev:
I think it just needs some time before we integrate everything & cut the release. ETA: this week

PR #205, opened 2018-02-02
[JENKINS-41626] Fix path detection when subfolder older than root folder

Andne:
This is based on comments in the JIRA issue and some additional investigation.  I think I have the interaction with the `SCMHeadObserver` working correctly.

In the diagram in the JIRA issue, the directory containing Jenkinsfile is at an older revision than the root directory of the scan.  This results in subversion not resolving the path since a directory in the middle of the path tree didn't exist in that specific revision.  The simple fix it so just always scan using HEAD, however that means that any notifications to the `SCMHeadObserver` wouldn't contain a valid revision.  This fix scans using HEAD, and then once it finds the Jenkinsfile checks if the specific revision would have worked.  If it did, then it assumes that it needs to notify the observer using that revision.  If the final check fails, then it assumes that the branch is newly created and that it cannot notify using a specific revision since the checkout itself will probably fail (I didn't test that part).  In this case it notifies with the HEAD revision and assumes since the branch is new the specific revision doesn't matter.

### DIFF
--- a/src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
+++ b/src/main/java/jenkins/scm/impl/subversion/SubversionSCMSource.java
@@ -384,7 +384,7 @@ public class SubversionSCMSource extends SCMSource {
                             final long candidateRevision = svnEntry.getRevision();
                             final long lastModified = svnEntry.getLastModified();
                             listener.getLogger().println(
-                                    "Checking candidate branch " + candidateRootPath + "@" + candidateRevision);
+                                    "Checking candidate branch " + candidateRootPath + "@HEAD");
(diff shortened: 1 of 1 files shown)


imakowski:
When this will be released?

oleg-nenashev:
Soon

On Mar 4, 2018 21:39, "imakowski" <notifications@github.com> wrote:

When this will be released?

—
You are receiving this because your review was requested.

Reply to this email directly, view it on GitHub
<https://github.com/jenkinsci/subversion-plugin/pull/205#issuecomment-370261963>,
or mute the thread

imakowski:
👍

PR #206, opened 2018-02-12
[JENKINS-49624] Fix selecting credentials for Pipeline Libraries using Modern SCM

Andne:
This is similar to JENKINS-44271, applies the same fix that was applied to the git plugin for that issue. I also found a conflict with passing the parameter to a couple other functions that didn't appear to make use of them.  As best as I can tell, nothing is using these functions so I think removing the unused parameter should be ok.


oleg-nenashev:
FTR https://issues.jenkins-ci.org/browse/JENKINS-44271

Andne:
JIRA ticket created: https://issues.jenkins-ci.org/browse/JENKINS-49624

oleg-nenashev:
Thanks @Andne!

PR #207, opened 2018-02-19
svnkit 1.9.1

ssl-maxtouch:
SVNKit 1.9.1 passes the unit tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,12 +80,12 @@ THE SOFTWARE.
     <dependency>
       <groupId>org.tmatesoft.svnkit</groupId>
       <artifactId>svnkit</artifactId>
-      <version>1.8.14</version>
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Is there any particular justification for the update (bugs in Jenkins JIRA, new features, etc.)? Just interested

ssl-maxtouch:
Yes, we found that 1.9 is better in merges and resolving conflicts, in fact we all use locally TortoiseSVN 1.9.x and it would be great if Jenkins would also use that (if there are no reasons to stick with the old version) without us needing to rebuild the plugin ourselves

oleg-nenashev:
FYI @kuisathaverat

imakowski:
When this will be released?

PR #211, opened 2018-03-22
[JENKINS-50339] - Whitelist SVNErrorMessage and introduce RemotableSVNErrorMessage

oleg-nenashev:
- [x] - Model objects of SVNErrorMessage are now whitelisted. It should work as long as there is no unsupported exception classes or objects in `Object[] myObjects`
- [x] - Introduce new RemotableSVNErrorMessage, which does not offer dangerous constructors && wraps exceptions into `ProxyException`. 
- [x] - Migrate all logic in the plugin to RemotableSVNErrorMessage
- [x] - Test 2.107.1 in ci.jenkins.io

Maybe RemotableSVNErrorMessage should check whether the constructor runs on the master and use ProxyException only on the agent side.

https://issues.jenkins-ci.org/browse/JENKINS-50339

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin()
+buildPlugin(jenkinsVersions: [null, "2.107.1"])
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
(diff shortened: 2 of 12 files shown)


oleg-nenashev:
@reviewbybees

oleg-nenashev:
[WARNING] Rule 3: org.apache.maven.plugins.enforcer.EnforceBytecodeVersion failed with message:

Found Banned Dependency: org.jenkins-ci:task-reactor:jar:1.5

Okay

PR #212, opened 2018-03-27
[FIX JENKINS-45801] and [FIX JENKINS-49219], when calculate change log of externals, add module location's credentials as default credentials

Liyaa:
when calculate change log of externals add module location's credentials as default credentials, fix [JENKINS-45801] and [JENKINS-49219]

### DIFF
--- a/src/main/java/hudson/scm/SubversionChangeLogBuilder.java
+++ b/src/main/java/hudson/scm/SubversionChangeLogBuilder.java
@@ -55,8 +55,8 @@ import java.io.IOException;
 import java.io.PrintStream;
 import java.io.File;
 import java.io.Serializable;
+import java.util.List;
 import java.util.Map;
-import java.util.Collection;
 import javax.annotation.Nonnull;
 import jenkins.MasterToSlaveFileCallable;
 
@@ -101,7 +101,7 @@ public final class SubversionChangeLogBuilder {
         this.env = env;
     }
 
-    public boolean run(Collection<SubversionSCM.External> externals, Result changeLog) throws IOException, InterruptedException {
+    public boolean run(Map<String, List<SubversionSCM.External>> externalsMap, Result changeLog) throws IOException, InterruptedException {
         boolean changelogFileCreated = false;
 
         TransformerHandler th = createTransformerHandler();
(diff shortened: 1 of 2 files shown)


swidbsk:
This seems to have fixed spurious authentication issues on repos with lots of externals for us

samitok:
This solves our problem with externals directing in the same repository. I also confirm.

src/main/java/hudson/scm/SubversionChangeLogBuilder.java
```diff
@@ -101,7 +101,7 @@ public SubversionChangeLogBuilder(Run<?,?> build, FilePath workspace, @Nonnull S
         this.env = env;
     }
 
-    public boolean run(Collection<SubversionSCM.External> externals, Result changeLog) throws IOException, InterruptedException {
+    public boolean run(Map<String, List<SubversionSCM.External>> externalsMap, Result changeLog) throws IOException, InterruptedException {
```
kuisathaverat: externalsMap should be @nonnull anotated

src/main/java/hudson/scm/SubversionChangeLogBuilder.java
```diff
@@ -101,7 +101,7 @@ public SubversionChangeLogBuilder(Run<?,?> build, FilePath workspace, @Nonnull S
         this.env = env;
     }
 
-    public boolean run(Collection<SubversionSCM.External> externals, Result changeLog) throws IOException, InterruptedException {
+    public boolean run(@Nonnull Map<String, List<SubversionSCM.External>> externalsMap, Result changeLog) throws IOException, InterruptedException {
```
dwnusbaum: This introduces a binary incompatibility, but [it doesn't look like any other plugins call this method](https://github.com/search?utf8=%E2%9C%93&q=user%3Ajenkinsci+SubversionChangeLogBuilder&type=Code). Might be good to add `@Restricted(NoExternalUse.class)` to the class in a followup PR.

src/main/java/hudson/scm/SubversionChangeLogBuilder.java
```diff
@@ -122,24 +122,20 @@ public boolean run(Collection<SubversionSCM.External> externals, Result changeLo
...
                 manager.dispose();
             }
         }
-        ISVNAuthenticationProvider authProvider =
-                CredentialsSVNAuthenticationProviderImpl
-                        .createAuthenticationProvider(build.getParent(), scm, null, listener);
-        final SVNClientManager manager = SubversionSCM.createClientManager(authProvider).getCore();
-        try {
-            SVNLogClient svnlc = manager.getLogClient();
-            for(SubversionSCM.External ext : externals) {
```
dwnusbaum: Is there any case in which this is still necessary as a fallback?

src/main/java/hudson/scm/SubversionChangeLogBuilder.java
kuisathaverat: I will manage it

src/main/java/hudson/scm/SubversionChangeLogBuilder.java
kuisathaverat: before of this change, there was only one list of remotes now are classified by remote, so now only process external when there are externals, the removed login it is not needed anymore.

dwnusbaum: Ok, seems fine to me then.

stefanbesler:
Hi!
Lately I am getting loads of

org.tmatesoft.svn.core.SVNCancelException: svn: E200015: E200015: ISVNAuthentication provider did not provide credentials; HTTP authorization 

exceptions (randomly ~ second build) and I hope that this commit solves this issue as I am dealing with externals on the same server too. @kuisathaverat when are you planning to release a new version containing this fix?

PR #213, opened 2018-04-04
Minor enhancements in readability

StefanSpieker:
I would like to participate in this plugin and I thought I could start by improving some unit tests. So this is just a very simple pull request to kindly ask if improving the unit tests is a good starting point at all.

### DIFF
--- a/src/main/resources/hudson/scm/browsers/FishEyeSVN/help-url_de.html
+++ b/src/main/resources/hudson/scm/browsers/FishEyeSVN/help-url_de.html
@@ -1,4 +1,4 @@
 <div>
   Geben Sie die FishEye-URL des Moduls an, z.B.
-  <tt>http://fisheye6.cenqua.com/browse/ant/<tt>.
+  <tt>http://fisheye6.cenqua.com/browse/ant/</tt>.
 </div>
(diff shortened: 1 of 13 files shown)


src/main/resources/hudson/scm/browsers/FishEyeSVN/help-url_de.html
```diff
@@ -1,4 +1,4 @@
 <div>
   Geben Sie die FishEye-URL des Moduls an, z.B.
-  <tt>http://fisheye6.cenqua.com/browse/ant/<tt>.
+  <tt>http://fisheye6.cenqua.com/browse/ant/</tt>.
```
oleg-nenashev: oops

StefanSpieker:
I don't understand why the build failed?

dwnusbaum:
This plugin is targeting [Java 7](https://github.com/jenkinsci/subversion-plugin/blob/776e7758f35262696f3b5c3b9bfe270b80c95f09/pom.xml#L139), so the lambdas need to be reverted to anonymous classes (unless the maintainers want to target Java 8).

dwnusbaum:
Looks like #214 will bump java.level to 8.

kuisathaverat:
yep, this Update core to 2.60 so needs java 8

oleg-nenashev:
https://github.com/jenkinsci/subversion-plugin/pull/216

PR #215, opened 2018-04-18
[JENKINS-50287] allow to select two different behaviours for as-it-is checkout

kuisathaverat:
[JENKINS-50287](https://issues.jenkins-ci.org/browse/JENKINS-50287)

@reviewbybees

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -2975,8 +2975,10 @@ public class SubversionSCM extends SCM implements Serializable {
          * @return {@link org.tmatesoft.svn.core.SVNDepth} value.
          */
         public SVNDepth getSvnDepthForCheckout() {
-            if(getDepthOption().equals("unknown")) {
+            if("unknown".equals(getDepthOption())) {
                 return SVNDepth.FILES;
+            } else if ("as-it-is-infinity".equals(getDepthOption())){
+                return SVNDepth.INFINITY;
             } else {
                 return getSvnDepth(getDepthOption());
             }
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -2975,8 +2975,10 @@ public SVNDepth getSvnDepthForUpdate() {
          * @return {@link org.tmatesoft.svn.core.SVNDepth} value.
          */
         public SVNDepth getSvnDepthForCheckout() {
-            if(getDepthOption().equals("unknown")) {
+            if("unknown".equals(getDepthOption())) {
                 return SVNDepth.FILES;
+            } else if ("as-it-is-infinity".equals(getDepthOption())){
```
dwnusbaum: Is this covered by existing tests? If not it would be great to add a new one.

src/test/java/jenkins/scm/impl/subversion/SubversionStepTest.java
```diff
@@ -116,4 +116,52 @@ public void multipleSCMs() throws Exception {
         assertFalse(iterator.hasNext());
     }
 
+
+    @Test
+    public void checkoutDepthAsItIsInfiniteTest() throws Exception {
+        checkoutDepthTest("as-it-is-infinity");
```
dwnusbaum: These tests check that the parameter works, but they don't test how the behaviors differ. Is it easy/possible to use `sampleRepo` to construct different repo histories so you can verify the different behavior of each parameter?

dwnusbaum: After quick discussion it seems this is not so easy, so I am ok with just the basic tests.

PR #216, opened 2018-05-02
java 8 required for presence of lambdas and svnkit to 1.9.2

ssl-maxtouch:
It is currently impossible to build because there are lambdas in recent commits that require java 8. We are also using svnkit 1.9.2 without problems and all unit tests pass

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -135,8 +135,8 @@ THE SOFTWARE.
   </scm>
 
   <properties>
-    <jenkins.version>2.7.3</jenkins.version>
-    <java.level>7</java.level>
+    <jenkins.version>2.54</jenkins.version>
+    <java.level>8</java.level>
     <no-test-jar>false</no-test-jar>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -135,8 +135,8 @@ THE SOFTWARE.
   </scm>
 
   <properties>
-    <jenkins.version>2.7.3</jenkins.version>
-    <java.level>7</java.level>
+    <jenkins.version>2.54</jenkins.version>
```
oleg-nenashev: Better to use 2.60.x LTS then

ssl-maxtouch: I changed to the minimum version requiring java 8

dwnusbaum: I think Oleg's point is that it is generally preferred to use LTS versions for baselines, and 2.60 is the first LTS that required Java 8.

ssl-maxtouch: I will change to the LTS then

PR #219, opened 2018-06-08
JENKINS-51817 fix NPEs

BerndFarkaDyna:
this PR fixes NPEs thorwn descriped in JENKINS-51817 due to wrong usage of SVNErrorMessage C'tors

### DIFF
--- a/src/main/java/jenkins/scm/impl/subversion/RemotableSVNErrorMessage.java
+++ b/src/main/java/jenkins/scm/impl/subversion/RemotableSVNErrorMessage.java
@@ -36,18 +36,18 @@ import org.tmatesoft.svn.core.SVNErrorMessage;
 public class RemotableSVNErrorMessage extends SVNErrorMessage {
 
     public RemotableSVNErrorMessage(SVNErrorCode code) {
-        super(code, null, null, null, 0);
+        super(code, null, new Object[]{}, null, 0);
     }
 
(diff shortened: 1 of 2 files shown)


kuisathaverat:
taking a look at the create method seems like message it is beter to be not null too, Could yo make the changes to fill message with and empty String if it is null? then I merged it directly.

```
	/**
	 * Creates an error message given an error code, description, an error type
	 * (whether it's a warning or an error) and may be related objects to be
	 * formatted with the error description and an optional cause.
	 * To format the provided <code>objects</code>
	 * with the <code>message</code>, you should use valid format patterns parsable for
	 * {@link MessageFormat}.
	 *
	 * @param  code       an error code
	 * @param  message    an error description
	 * @param  objects    an array of objects related to the error <code>message</code>
	 * @param  type       an error type
	 * @param  cause     cause of the error
	 * @return            a new error message
	 */
	public static SVNErrorMessage create(SVNErrorCode code, String message, Object[] objects, int type, Throwable cause) {
	    return new SVNErrorMessage(code == null ? SVNErrorCode.BASE : code, message == null ? "" : message,
	            objects == null ? EMPTY_ARRAY : objects, cause, type);
	}
```

BerndFarkaDyna:
@kuisathaverat sure no problem I'll do the change

bfarka:
Is there an ETA when this fix will be available in an published version?

PR #226, opened 2018-10-05
Migrate into localization-zh-cn-plugin

LinuxSuRen:
See https://github.com/jenkinsci/localization-zh-cn-plugin


LinuxSuRen:
@kuisathaverat I wonder if you could help review this PR? Thanks.

PR #228, opened 2018-12-26
JENKINS-27079 ClassCastException: hudson.scm.SVNRevisionState cannot …

yanpak:
Class checking was provided.
See my comment https://issues.jenkins-ci.org/browse/JENKINS-27079?focusedCommentId=356638&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-356638

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -752,9 +752,11 @@ public class SubversionSCM extends SCM implements Serializable {
         // maybe some XSLT engine doesn't close the stream properly.
         // so let's do it by ourselves to be really sure that the stream gets closed.
         OutputStream os = new BufferedOutputStream(new FileOutputStream(changelogFile));
-        boolean created;
+        boolean created = false;
         try {
-            created = new SubversionChangeLogBuilder(build, workspace, (SVNRevisionState) baseline, env, listener, this).run(externalsMap, new StreamResult(os));
+            if(baseline instanceof SVNRevisionState){
+                created = new SubversionChangeLogBuilder(build, workspace, (SVNRevisionState) baseline, env, listener, this).run(externalsMap, new StreamResult(os));
+            }
         } finally {
             os.close();
         }


src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -752,9 +752,11 @@ private void calcChangeLog(Run<?,?> build, FilePath workspace, File changelogFil
         // maybe some XSLT engine doesn't close the stream properly.
         // so let's do it by ourselves to be really sure that the stream gets closed.
         OutputStream os = new BufferedOutputStream(new FileOutputStream(changelogFile));
-        boolean created;
+        boolean created = false;
         try {
-            created = new SubversionChangeLogBuilder(build, workspace, (SVNRevisionState) baseline, env, listener, this).run(externalsMap, new StreamResult(os));
+            if(baseline instanceof SVNRevisionState){
+                created = new SubversionChangeLogBuilder(build, workspace, (SVNRevisionState) baseline, env, listener, this).run(externalsMap, new StreamResult(os));
+            }
```
oleg-nenashev: else print warning (e.g. to the system log)?

yanpak: I think, logging will be useful only for debugin purpoises. I will make it in my own environment to find the source of ClassCastException problem.
If I find (soon or later I cannot predict from this point) I will make new PR.
Now, my current PR will save many nerves for users of subversion-plugin :) Hope, you agree with me

oleg-nenashev: I agree for sure, it is not a blocker from my PoV. I just approved it with a suggestion

yanpak: Thank you, Oleg!

README.md
```diff
@@ -6,3 +6,6 @@ Provides Jenkins integration with [Apache Subversion](http://subversion.apache.o
 See the [Subversion Plugin page](https://plugins.jenkins.io/subversion) for more information.
 
 Apache, Apache Subversion and Subversion are trademarks of the Apache Software Foundation.
+
+> Here is fixed build crushing with error:<br/>
+> <i>hudson.scm.SVNRevisionState cannot be cast to hudson.scm.SVNRevisionState</i>
```
jglick: You can revert the change to the readme.

PR #229, opened 2019-02-08
Support jenkins configuration as code plugin

cohencil:
Jenkins CasC configuration snippet:
```
subversionSCM:
  globalExcludedRevprop: 'globalExcludedRevprop'
  workspaceFormat: 8
```

`workspaceFormat` to Subversion version mapping:

workspaceFormat | version
-- | --
8 | 1.4
9 | 1.5
10 | 1.6
29 | 1.7
31 | 1.8

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -2170,7 +2170,6 @@ public class SubversionSCM extends SCM implements Serializable {
             return "Subversion";
         }
 
-        @Restricted(NoExternalUse.class)
         public void setGlobalExcludedRevprop(String globalExcludedRevprop) {
             this.globalExcludedRevprop = globalExcludedRevprop;
         }
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -2171,18 +2171,22 @@ public class SubversionSCM extends SCM implements Serializable {
         }
 
         @Restricted(NoExternalUse.class)
-        void setGlobalExcludedRevprop(String revprop) {
-            globalExcludedRevprop = revprop;
+        public void setGlobalExcludedRevprop(String globalExcludedRevprop) {
+            this.globalExcludedRevprop = globalExcludedRevprop;
         }
 
         public String getGlobalExcludedRevprop() {
-            return globalExcludedRevprop;
+            return this.globalExcludedRevprop;
         }
 
         public int getWorkspaceFormat() {
-            if (workspaceFormat==0)
+            if (workspaceFormat == 0)
                 return SVNAdminAreaFactory.WC_FORMAT_14; // default
             return workspaceFormat;
+        }
+
+        public void setWorkspaceFormat(int workspaceFormat) {
+            this.workspaceFormat = workspaceFormat;
         }
 
         public boolean isValidateRemoteUpToVar() {


yorammi:
For not confusing the users, I suggest something like this (I didn't test it still):

>          public int getWorkspaceFormat() {
>             if (workspaceFormat==null ) {
>                 return SVNAdminAreaFactory.WC_FORMAT_14
>             }
>             return (workspaceFormat=='1.4'?
>                         SVNAdminAreaFactory.WC_FORMAT_8
>                         :(workspaceFormat=='1.5'?
>                          SVNAdminAreaFactory.WC_FORMAT_9
>                          :(workspaceFormat=='1.6'?
>                           SVNAdminAreaFactory.WC_FORMAT_10
>                           :(workspaceFormat=='1.7'?
>                            SVNAdminAreaFactory.WC_FORMAT_29
>                            :(workspaceFormat=='1.8'?
>                             SVNAdminAreaFactory.WC_FORMAT_31
>                             :SVNAdminAreaFactory.WC_FORMAT_14)))))
>         }

yorammi:
Anyone is releasing new versions of this plugin? The last version is from Sep.2018 after commits from @kuisathaverat (I think). I can't really use a locally build of the plugin since I'm automatic the servers plugins list (using plugins.txt file and CasC)

kuisathaverat:
I have stoped maintaining that plugin months ago, I don’t have time or interest on it any more.

yorammi:
Who can take it and merge @cohencil PR to release a new version? @kohsuke, what is the procedure if no one is the owner and one (I can do it if needed) wants to take it?
I need this plugin for a customer project (they still working with SVN even that I'm in a process of moving them to Git, but it will take few more months)

BlueAndi:
@cohencil @yorammi Is there a workaround to set the workspace format, until the plugin gets updated?

cohencil:
@BlueAndi you can always build this PR locally and use the method above to set the workspace format.

BlueAndi:
@cohencil Unfortunately I am not so familiar with the toolchain, so I feared this answer. :-) But I will try.

BlueAndi:
@cohencil We build the plugin now (yipieeh :-) ), but we have trouble with the permissions during building the image.

```
COPY plugins.txt /usr/share/jenkins/ref/plugins.txt
COPY ./plugins/subversion-2.12.2-SNAPSHOT.jpi /var/jenkins_home/plugins/
RUN chgrp jenkins /var/jenkins_home/plugins/subversion-2.12.2-SNAPSHOT.jpi
RUN chown jenkins /var/jenkins_home/plugins/subversion-2.12.2-SNAPSHOT.jpi
RUN /usr/local/bin/install-plugins.sh < /usr/share/jenkins/ref/plugins.txt
```
We get the error:

> chgrp: changing group of '/var/jenkins_home/plugins/subversion-2.12.2-SNAPSHOT.jpi': Operation not permitted

Could you help again?

kuisathaverat:
If what you want it is a binary with this fix you could take it from the Jenkins build at https://ci.jenkins.io/job/Plugins/job/subversion-plugin/view/change-requests/job/PR-229/

BlueAndi:
@kuisathaverat Thanks, for the info! Somehow we were not aware of that we could use the build PR binary from the Jenkins build.

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -2171,20 +2171,24 @@ public String getDisplayName() {
...
         }
 
         public String getGlobalExcludedRevprop() {
-            return globalExcludedRevprop;
+            return this.globalExcludedRevprop;
         }
 
         public int getWorkspaceFormat() {
-            if (workspaceFormat==0)
+            if (workspaceFormat == 0)
```
jglick: Please avoid touching unrelated code, especially for reformatting.

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -2171,20 +2171,24 @@ public String getDisplayName() {
         }
 
         @Restricted(NoExternalUse.class)
-        void setGlobalExcludedRevprop(String revprop) {
-            globalExcludedRevprop = revprop;
+        public void setGlobalExcludedRevprop(String globalExcludedRevprop) {
```
oleg-nenashev: It will not (or should not) work with `@Restricted` without extra settings

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -2171,20 +2171,24 @@ public String getDisplayName() {
...
         }
 
         public int getWorkspaceFormat() {
-            if (workspaceFormat==0)
+            if (workspaceFormat == 0)
                 return SVNAdminAreaFactory.WC_FORMAT_14; // default
             return workspaceFormat;
         }
 
+        public void setWorkspaceFormat(int workspaceFormat) {
```
oleg-nenashev: Nice 2 have: Some javadoc

oleg-nenashev:
CC @jenkinsci/code-reviewers

JungStar:
We would be really interested in the JCasc support. Is there a chance to get it merged and released in the near future?

src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -2171,20 +2171,24 @@ public String getDisplayName() {
         }
 
         @Restricted(NoExternalUse.class)
```
timja, line 2173:
```suggestion
```

PR #233, opened 2019-04-29
Hyphenated username

bryceschober:
Allow hyphens in excluded usernames


src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -2463,7 +2463,7 @@ public FormValidation doCheckIncludedRegions(@QueryParameter String value) throw
          * Regular expression for matching one username. Matches 'windows' names ('DOMAIN&#92;user') and
          * 'normal' names ('user'). Where user (and DOMAIN) has one or more characters in 'a-zA-Z_0-9')
```
jglick: Javadoc is no longer accurate.

bryceschober: Fixed.

PR #238, opened 2019-07-26
Fix FreeBSD12 compatibility problem

spagu:
https://svn.svnkit.com/repos/svnkit/tags/1.10.1/CHANGES.txt
+ OSGi bundle (eclipse update site) fixed to include LZ4 library classes.
+ FreeBSD compatibility fixes.
+ Possible repository corruption when using SVNKit DAV servlet fixed.
+ Java 11 support
+ SVN 1.10 repository format support

Reason for that is this subversion-plugin does not work on FreeBSD12

https://svnkit.com/download.php ( added support only for FreeBSD12-CURRENT hardcoded )
https://issues.tmatesoft.com/issue/SVNKIT-740
https://svn.svnkit.com/repos/svnkit/tags/1.10.1/svnkit/src/main/java/org/tmatesoft/svn/core/internal/wc/SVNFileUtil.java


oleg-nenashev:
Cc @jenkinsci/java11-support since it is mentioned in the list

pom.xml
```diff
@@ -173,7 +173,7 @@ THE SOFTWARE.
     <dependency>
       <groupId>org.tmatesoft.svnkit</groupId>
       <artifactId>svnkit-cli</artifactId>
-      <version>1.9.3</version>
+      <version>1.10.1</version>
```
jglick: We should be introducing a POM property, then enabling Dependabot.

PR #244, opened 2020-02-01
Replace "Hudson" with "Jenkins"

darxriggs:


darxriggs:
No. Sorry. I just improve things here and there in some plugins.

PR #247, opened 2020-11-04
[JENKINS-47803 && JENKINS-7009] Remove working copy if external is broken

PhilippHomann:
Currently checkout silently ignores an external failure due to locked working copy.
So build continues without one recognizing that the result is build using an old revision.

With this fix the update will behave like the non-external checkout: delete working copy and start with a fresh one.


PhilippHomann:
Anyone?
@daniel-beck @jglick

daniel-beck:
I'm not a maintainer. The only reason I show up in the history is because I staged security fixes.

jglick:
I also do not pretend to maintain this plugin. Not sure if anyone does.

src/test/java/hudson/scm/SubversionSCMTest.java
```diff
@@ -1947,4 +1952,80 @@ private static File newFolder(File root, String... subDirs) throws IOException {
...
+        Connection conn = DriverManager.getConnection("jdbc:sqlite:" + dbFile);
+        PreparedStatement stmt = conn.prepareStatement("INSERT INTO WC_LOCK VALUES(1, ?, 1)");
+        stmt.setString(1, pathToLock);
+        stmt.executeUpdate();
+        conn.close();
+    }
+
+    @Issue("JENKINS-47803")
+    @Test
+    public void lockedExternal() throws Exception {
```
Eccenux, line 1967: should have a comment on the setup phase of the test – what is being setup and why

src/test/java/hudson/scm/SubversionSCMTest.java
```diff
@@ -1947,4 +1952,80 @@ private static File newFolder(File root, String... subDirs) throws IOException {
...
+            r.assertLogContains("Workspace appear to be locked, so getting a fresh workspace", build);
+        } finally {
+            p.kill();
+        }
+    }
+
+    @Test
+    public void ensureCancelOnExternalsFail() throws Exception {
+        Proc p = runSvnServe(getClass().getResource("JENKINS-777.zip"));
+
```
Eccenux, line 2002: Same as above. should have a comment on the setup phase of the test – what is being setup and why

src/test/java/hudson/scm/SubversionSCMTest.java
```diff
@@ -1947,4 +1952,80 @@ private static File newFolder(File root, String... subDirs) throws IOException {
         }
         return result;
     }
-}
+
+    void lockWorkspace(String dbFile, String pathToLock) throws Exception {
```
Eccenux, line 1956: That is an interesting way to emulate a lock. Is this documented somewhere?

Eccenux (commented):
I'm confused about this change:
https://github.com/jenkinsci/subversion-plugin/commit/b04c6c2a4552989a98c6e21fab5a1a3bd2319c47

Is that related to detecting broken externals?

PR #248, opened 2021-01-11
subversion external handling

theKBro:
Solving some problems with externals 

1. same external linked to projekt in different revisions
2. unknown credential for externals on polling

Should also solve: JENKINS-22542

### DIFF
--- a/src/main/java/hudson/scm/CompareAgainstBaselineCallable.java
+++ b/src/main/java/hudson/scm/CompareAgainstBaselineCallable.java
@@ -117,7 +117,15 @@ final class CompareAgainstBaselineCallable extends MasterToSlaveCallable<Polling
         changes.changes |= (nowRev>baseRev);
 
         listener.getLogger().println(Messages.SubversionSCM_pollChanges_remoteRevisionAt(url, nowRev));
-        revs.put(url, nowRev);
+        if(revs.containsKey(url)){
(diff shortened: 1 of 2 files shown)


theKBro:
anyone with write access?
@timja @StefanSpieker

timja:
Sorry I have no way of testing this or can understand it really,

I've just shipped some maintenance fixes to allow tests to run on Java 11 basically.

If you want this shipped the best way would be to request maintainer access as by this PR I assume you actually have an SVN environment and are actively using it in your organisation.

You can add yourself to https://github.com/jenkins-infra/repository-permissions-updater/blob/master/permissions/plugin-subversion.yml with your Jenkins project account (the one you would use for Jira), (if you want to do this).

theKBro:
@timja 

> If you want this shipped the best way would be to request maintainer access as by this PR I assume you actually have an SVN environment and are actively using it in your organisation.

Yes, we already tested the changes in our production system for a while now. So how to request maintainer access?

> You can add yourself to https://github.com/jenkins-infra/repository-permissions-updater/blob/master/permissions/plugin-subversion.yml with your Jenkins project account (the one you would use for Jira), (if you want to do this).

Would be great but this is also write protected for me.

timja:
Just send a pull request there adding yourself to that list

PR #251, opened 2021-02-08
replaced deprecated jenkins.getInstance with replacements and minor cleanup

StefanSpieker:
replaced deprecated `Jenkins.getInstance()` with replacements and minor cleanup


Eccenux:
Oh, sorry, I didn't notice this. I think I already did all that in: https://github.com/jenkinsci/subversion-plugin/commit/2f09f9988915712c51c9fbbad7adfd80e95996a2

PR #254, opened 2021-03-24
[JENKINS-65161] remove commons-digester2 from core and upgrade plugins to commons-digester3

olamy:


olamy:
https://github.com/jenkinsci/jenkins/pull/5320

rsandell (approved):

olamy:
@jglick or @timja  ping (as you are the maintainer with most of the commit here)

timja:
i've retriggered the build

jglick:
Or @theKBro as per #1807?

jglick:
Windows CI failing. Maybe best to disable it.

pom.xml
```diff
@@ -176,6 +176,12 @@ THE SOFTWARE.
         <artifactId>json</artifactId>
         <version>20190722</version>
       </dependency>
+      <dependency>
+        <groupId>commons-beanutils</groupId>
+        <artifactId>commons-beanutils</artifactId>
+        <!-- 1.9.4 when this merged https://github.com/stapler/stapler/pull/211 -->
```
daniel-beck, line 182: Why wait?

theKBro: I also don't understand this. Is this needed by common-digister3?

olamy: nope can be removed. commons-beanutils 1.9.4 is now supported by core

basil:
@timja @jglick @olamy I fear something may be broken as a result of this recent work ... maybe not in production, but at least in integration tests.

Try running `mvn test -Djenkins.version=2.297 -Dtest=org.jenkinsci.plugins.workflow.steps.scm.SCMStepTest#checkoutsRestored` in `workflow-scm-step`. It passes on current `master`. But, try changing `pom.xml` with

```
diff --git a/pom.xml b/pom.xml
index 0f487eb..836bb07 100644
--- a/pom.xml
+++ b/pom.xml
@@ -42,7 +42,7 @@
         <changelist>-SNAPSHOT</changelist>
         <jenkins.version>2.222.4</jenkins.version>
         <java.level>8</java.level>
-        <subversion-plugin.version>2.14.0</subversion-plugin.version>
+        <subversion-plugin.version>2.14.1</subversion-plugin.version>
         <svnkit.version>1.10.1</svnkit.version>
     </properties>
     <dependencyManagement>
```

Now the test fails with

```
java.lang.ArrayIndexOutOfBoundsException: 153981
        at org.objectweb.asm.ClassReader.<init>(Unknown Source)
        at org.objectweb.asm.ClassReader.<init>(Unknown Source)
        at org.objectweb.asm.ClassReader.<init>(Unknown Source)
        at org.kohsuke.stapler.ClassDescriptor$ASM.loadParametersFromAsm(ClassDescriptor.java:312)
        at org.kohsuke.stapler.ClassDescriptor$ASM.access$000(ClassDescriptor.java:300)
        at org.kohsuke.stapler.ClassDescriptor.loadParameterNames(ClassDescriptor.java:196)
        at org.kohsuke.stapler.Function$MethodFunction.getParameterNames(Function.java:356)
        at org.kohsuke.stapler.Function.bindAndInvoke(Function.java:173)
        at org.kohsuke.stapler.MetaClass$9.dispatch(MetaClass.java:455)
        at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:766)
        at org.kohsuke.stapler.Stapler.invoke(Stapler.java:898)
        at org.kohsuke.stapler.Stapler.invoke(Stapler.java:694)
        at org.kohsuke.stapler.Stapler.service(Stapler.java:240)
```

Interestingly enough, if you keep Subversion 2.14.1 but try again with `mvn test -Djenkins.version=2.295 -Dtest=org.jenkinsci.plugins.workflow.steps.scm.SCMStepTest#checkoutsRestored`, the test passes. So core 2.295 works with Subversion 2.14.0 and 2.14.1, but core 2.296 and 2.297 work with Subversion 2.14.0 but not 2.14.1. Subversion 2.14.2 is similarly broken, but only on core 2.296 and 2.297.

Suppose you revert commit 628e74896ef4fa80cf0929f81ac69a8ce5e469cf from `master` of Subversion. That yields a `2.14.3-SNAPSHOT`. Now try picking that up in `workflow-scm-step`. Behold, `mvn test -Djenkins.version=2.297 -Dtest=org.jenkinsci.plugins.workflow.steps.scm.SCMStepTest#checkoutsRestored` now passes!

I took a look at the difference between `subversion.hpi` with commit 628e74896ef4fa80cf0929f81ac69a8ce5e469cf reverted and not reverted. It seems that commit 628e74896ef4fa80cf0929f81ac69a8ce5e469cf added to this JAR `commons-digester3-3.2.jar`, `commons-beanutils-1.9.3.jar`, `cglib-2.2.2.jar`, and `asm-3.3.1.jar`. That's about as far as I got in investigating this. Something seems off, but I can't tell if it's a production issue or just a test classpath issue.

basil:
OK, I think I figured out what's going on here. By changing Subversion to

```
diff --git a/pom.xml b/pom.xml
index ac012e8..77fae12 100644
--- a/pom.xml
+++ b/pom.xml
@@ -214,6 +214,12 @@ THE SOFTWARE.
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-digester3</artifactId>
       <version>3.2</version>
+      <exclusions>
+        <exclusion>
+          <groupId>asm</groupId>
+          <artifactId>asm</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
```

and pulling that in via `workflow-scm-step`, `mvn test -Djenkins.version=2.297 -Dtest=org.jenkinsci.plugins.workflow.steps.scm.SCMStepTest#checkoutsRestored` now passes.

I think the problem is that the Subversion plugin is now bundling ASM, which conflicts with the ASM in Jenkins core. Subversion should either exclude ASM like I did above, or use `maskClasses` like `jacoco-plugin` does. Either way, the current behavior is wrong. This probably applies to the other Digester removal diffs as well.

Yes, my change in stapler/stapler#225 makes this more obvious by having Stapler use the regular ASM classes rather than a repackaged version, which explains why this is only visible starting in 2.296. Rolling that back would reduce the scope of the problem but wouldn't completely solve it... the same exception could occur with any other area of core that uses ASM (like JNR). The problem really is with the Digester removal PRs, which didn't exclude or mask ASM which is already provided by core.

jglick:
FYI @jtnord ^^^

jtnord:
well how timely :)  

https://groups.google.com/g/jenkinsci-dev/c/NhV_o6zxbzw

timja:
@jtnord core has included a non shaded ASM for a long time via jna

jtnord:
> @jtnord core has included a non shaded ASM for a long time via jna

*cough* and it has just been updated and it is not compatable :-)

jtnord:
regardless of the ASM upgrade whatever etc etc in core,  the updated plugins bundling beansutils is clearly undesired.  (and the use of an unshaded ASM is also likely an oversight)

jtnord:
https://commons.apache.org/proper/commons-digester/

image

That said - the set of dependencies looks quiesionable 😢

jtnord:
@basil thanks for the report, I filed [JENKINS-65862](https://issues.jenkins.io/browse/JENKINS-65862) and we are looking at this.

PR #257, opened 2021-06-11
[JENKINS-65862] exclude asm and common-beanutils from commons-digester3

car-roll:
See [JENKINS-65862](https://issues.jenkins.io/browse/JENKINS-65862)
When `commons-digester3` was added as a dependency, it bundled in `asm` and `common-beanutils`, both of which are included in core. This PR excludes `asm` and `common-beanutils` from `commons-digester3`

upstream of https://github.com/jenkinsci/workflow-scm-step-plugin/pull/53

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -214,6 +214,16 @@ THE SOFTWARE.
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-digester3</artifactId>
       <version>3.2</version>
+      <exclusions>
+        <exclusion>
+          <groupId>asm</groupId>
+          <artifactId>asm</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>commons-beanutils</groupId>
+          <artifactId>commons-beanutils</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>


basil (approved):

basil:
Hrm, the more I think about this, the more I wonder if I approved this too soon. Digester's ASM dependency on ASM 3 is via Cglib, which we are still shipping. So will Cglib, compiled against ASM 3, work against the ASM 9 in core? I guess it is in practice if `SubversionChangeLogParserTest`, but it does make me a bit nervous. Alternatively, we could revert this, keeping Subversion as-is, and just fix the tests by excluding ASM not in Subversion but in the 3 plugins that consume Subversion in test context. What do you think, @car-roll?

basil:
I guess Tim released this already so we might as well see what happens. I'm probably being overly paranoid, but at least if there are any problems we'll know what to do to fix them.

PR #258, opened 2021-06-11
Safer way of excluding ASM

basil:
Amends #257. As I have discovered on this saga, ASM 4 and later are good about compatibility, but ASM 3 and earlier are not. Unfortunately Digester depends on ASM 3 indirectly, through cglib:

```
[INFO] +- org.apache.commons:commons-digester3:jar:3.2:compile
[INFO] |  +- cglib:cglib:jar:2.2.2:compile
[INFO] |  |  \- asm:asm:jar:3.3.1:compile
[INFO] |  \- commons-beanutils:commons-beanutils:jar:1.9.3:compile
```

So by excluding ASM but still packaging cglib, as was done in #257, we are running cglib (compiled against 3.3.1) against whatever ASM is provided by core (9.1), which seems risky.

This change instead excludes _cglib_ and then adds `cglib-nodep`, which bundles its own repackaged version of ASM. This way Digester still gets what it needs (cglib 2.2.2) without worrying about any unnecessary risk.

CC @timja @car-roll

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -215,15 +215,24 @@ THE SOFTWARE.
       <artifactId>commons-digester3</artifactId>
       <version>3.2</version>
       <exclusions>
+        <!--
+          cglib depends on ASM 3, which is incompatible with ASM 4 and later. Core includes ASM 9.
+          Exclude it here so that we can pull in a version with shaded ASM below.
+        -->
         <exclusion>
-          <groupId>asm</groupId>
-          <artifactId>asm</artifactId>
+          <groupId>cglib</groupId>
+          <artifactId>cglib</artifactId>
         </exclusion>
         <exclusion>
           <groupId>commons-beanutils</groupId>
           <artifactId>commons-beanutils</artifactId>
         </exclusion>
       </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>cglib</groupId>
+      <artifactId>cglib-nodep</artifactId>
+      <version>2.2.2</version>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>


basil:
So to summarize, I think there are several options:

- Exclude ASM and BeanUtils as was done in #257 and released in 2.14.3
- Replace cglib with a shaded version as is proposed in this PR. Cglib brings its own shaded ASM. Continue excluding BeanUtils to get it from core.
- Replace Digester with the `no-deps` classifier version. This shades _all_ of Digester's dependencies, including cglib and BeanUtils.
- Revert #257 and work around the test problem in tests by excluding ASM at the consumer side rather than the producer side

basil:
These choices are in order from riskiest to safest. Unfortunately I think we already picked the riskiest option in #257, but it isn't too late to correct that if we want to go with a different option. I have no strong preference.

timja:
No real opinion, happy to go with this or another option

basil:
@car-roll How do you feel about reverting #257 and instead excluding ASM at the consumer side of each plugin that consumes Subversion in its test suite? That way we have the lowest risk of regression while still getting BOM/PCT tests working again. I think that is my favorite option as it allow us all to move onto other things. Whereas the route we are currently taking makes me a bit nervous that something will blow up in cglib over the coming weeks.

bitwiseman:
The problem with the full digester nodep path is that there is no way to mitigate security issues found in the deps (cough, beanutils).  I'm okay with taking the nodep cglib to guarantee compat doesn't break uf we absolutely need to, but... It is quite possible that compat is fine and this isn't needed.

bitwiseman:
@basil 
I've created https://github.com/jenkinsci/bom/pull/557 to test this plugin. Assuming that looks good, would you be willing to merge and release this? 

I'm looking to this plugin as the canary for the other plugins that were updated for the removal of commons-digester.  The dependency and exclusions as shown in this PR seem to strike a reasonable balance between stability and security that other plugins can follow.  Would you agree?

basil:
> I've created jenkinsci/bom#557 to test this plugin. Assuming that looks good, would you be willing to merge and release this?

@timja has the merge and release permissions, but from my perspective this is fine.

> I'm looking to this plugin as the canary for the other plugins that were updated for the removal of commons-digester. The dependency and exclusions as shown in this PR seem to strike a reasonable balance between stability and security that other plugins can follow. Would you agree?

I agree in principle. The main risk in the current release of Subversion is that Digester 3.2 would invoke cglib 2.2.2 (compiled against ASM 3.3.1), which would invoke ASM 9.1 from core. Since ASM is compatible between versions 4 and 9, but not necessarily 3 and 9, there could be an issue. This approach is a bit safer in that Digester would invoke the "fat" version of cglib 2.2.2, which has its own ASM 3.3.1 inlined. The downside to this approach is that it makes maintaining the POM file a bit harder, as one has to be careful not to upgrade cglib to latest (3.3.0) without also verifying that cglib 3.3.0 is binary compatible with cglib 2.2.2. A Dependabot exclude should probably be added to any repository that adopts this approach and uses Dependabot (Subversion does not).

If this is to become a model for other plugins rather than a short-term workaround, we would ideally figure out how to exercise the "Digester invoking cglib invoking ASM" code path and conclusively determine which approach is best. I took a quick look at the Digester 3.2 sources and saw that Digester only invokes cglib in `org.apache.commons.digester3.ObjectCreateRule`, which is used to implement the "Default constructor arguments" feature in Digester. The first thing I would try to do is determine whether we use this feature or not (I'm not sure right now). For a thorough test, I would start Jenkins with `java -jar jenkins.war` (not `JenkinsRule`), install the Subversion plugin, connect with a remote debugger, and set a breakpoint on wherever we invoke `Digester` from Subversion and on `ObjectCreateRule` in Digester. If we hit the first breakpoint but not the second, that would indicate that we aren't using the "Default constructor arguments" feature, so cglib (let alone ASM) isn't really needed by Digester -- in which case the approach currently taken by the released version of the Subversion plugin is fine (and this could be adopted as a model for other plugins, provided they also don't use the "Default constructor arguments" feature). But if the second breakpoint gets hit, that would mean we are using the "Default constructor arguments" feature, which means we would expect to see cglib invoking core's ASM (for the released version of Subversion) or the shaded ASM (for this PR). Either way we would validate that this PR isn't needed or that it conclusively works.

Unfortunately I haven't had the time yet to dig deeper into this level of testing.

timja:
Given the above analysis it sounds to me better that we don't hold us on to ancient versions if we don't need to?

i.e. only go ahead with this change if we need to?

> @timja has the merge and release permissions, but from my perspective this is fine.

Plugin is up for adoption, anyone is welcome to do it, I only have them because of 'emergency' tables to divs fixes.

bitwiseman:
@basil 
Thank you for that next level of analysis. 

@basil @timja 
Looking at the [list of affected plugins in JEP-231](https://github.com/jenkinsci/jep/tree/master/jep/231#affected-plugins-fixes-planned), I don't think it is worth digging very hard on this.  After this plugin and the `cvs` plugin the numbers don't really justify it.  I'll file new PRs to update their Digester3 dependencies look the released version and link to this PR for reference.  I'll do the same in the JEP and we can table any further work for now.  

Sound good?

timja:
to be clear did you want this merged and released or not? I can do so if needed

jglick:
> If this is to become a model for other plugins rather than a short-term workaround, we would ideally figure out how to exercise the "Digester invoking cglib invoking ASM" code path

IMO the proper fix is to remove the dependency on _any_ version of Digester, as in https://github.com/jenkinsci/mercurial-plugin/commit/84af58b08f80bb92792f7bc04a31487f3eeee95a.

basil:
> IMO the proper fix is to remove the dependency on _any_ version of Digester, as in [jenkinsci/mercurial-plugin@84af58b](https://github.com/jenkinsci/mercurial-plugin/commit/84af58b08f80bb92792f7bc04a31487f3eeee95a).

Of course but that is even _more_ work than what I described. So this just comes down to a matter of how far do we want to go and who is willing to do the work. My interest is just in getting the BOM tests to pass again, which they now should be, so I can't volunteer to go all the way on this one.

jglick:
> that is even _more_ work than what I described

Well, perhaps, but it is just some plain old Java programming, compared to all the ink that has been spilled in this and preceding PRs trying to make Digester & ASM work. Up to the maintainers of this plugin, if there are any.

bitwiseman:
@timja 
> to be clear did you want this merged and released or not? I can do so if needed

No, I'm proposing we close this and keep it only for reference/in-case-needed.

bitwiseman:
> > that is even _more_ work than what I described
> 
> Well, perhaps, but it is just some plain old Java programming

It is also "just" time that none of us have that we would be spending on a plugin no one wants to own.

jglick:
#259 then. This lib should die.

PR #267, opened 2021-11-05
[JENKINS-48543] Remove agent-to-controller calls

jglick:
[JENKINS-48543](https://issues.jenkins.io/browse/JENKINS-48543)

Following up 7d1525edea6641a2febd3f7deeac55c0a89b0d7e to reduce risk and complexity, and provide compatibility with possible future hardenings such as https://github.com/jenkinsci/jenkins/pull/5885. Not tested beyond what automated test coverage exists here.

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -59,6 +59,7 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.BulkChange;
 import hudson.EnvVars;
 import hudson.Extension;
+import hudson.ExtensionList;
 import hudson.FilePath;
 import hudson.FilePath.FileCallable;
 import hudson.Launcher;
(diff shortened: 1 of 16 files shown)


src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -2692,8 +2692,7 @@ public static void init() {}
                 System.setProperty("svnkit.http.methods","Digest,Basic,NTLM,Negotiate");
 
             // use SVN1.4 compatible workspace by default.
-            SVNAdminAreaFactory.setSelector(new SubversionWorkspaceSelector());
-        }
+            SVNAdminAreaFactory.setSelector(new SubversionWorkspaceSelector(workspaceFormat));
```
jglick, line 2695: This part is tricky. SVNKit apparently has static state, which is a bad idea generally (forces us to do some work on both the Jenkins controller & agent JVMs) and is especially weird since there is a user-configurable format which we have to somehow pass into static state. Hope test coverage is good enough here.

jglick:
@didiez can you review this please? I can spend the time to fix up merge conflicts if someone is going to acknowledge that this exists.

didiez (approved):
Nice PR @jglick. Looks good to me :+1:
Sorry for the delay but I had too much on my plate.

PR #277, opened 2023-01-04
Enable CD

jglick:
Possible as of https://github.com/jenkins-infra/repository-permissions-updater/pull/3057. See: https://www.jenkins.io/doc/developer/publishing/releasing-cd/ and in particular the description of the version number format change.

@didiez this is just up to you; do not hesitate to close if you prefer to `mvn release:{prepare,perform}` manually with the existing version scheme. If this is merged, then it becomes possible to cut releases merely by merging appropriately-labeled PRs. (All PRs should be given one of the stock labels like `dependencies`, `chore`, `enhancement`, or `bug` before merge.)


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+version: 2
+updates:
+- package-ecosystem: maven
+  directory: /
+  schedule:
+    interval: monthly
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: monthly
```
jglick, line 12: Was not previously enabled; will produce a steady flow of PRs, though they can be ignored if you prefer. (Note that updating the POM would require a switch to 2.361.x and Java 11.)

pom.xml
```diff
@@ -48,86 +48,6 @@ THE SOFTWARE.
     </license>
   </licenses>
 
-  <developers>
```
jglick, line 51: All obsolete, not helpful.

timja (approved):

PR #279, opened 2023-04-27
Refresh plugin for April 2023

basil:
This is required for https://github.com/jenkinsci/stapler/pull/452 so that tests continue passing in this plugin and in BOM.


basil:
Used my admin powers to replay the build with the `Jenkinsfile` changes here: https://ci.jenkins.io/job/Plugins/job/subversion-plugin/job/PR-279/2/

PR #284, opened 2024-03-25
Fix missing dependency on directly used trilead-api

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

### Testing done

```[tasklist]
### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -68,6 +68,7 @@ import hudson.Util;
 import hudson.init.InitMilestone;
 
 import java.io.ByteArrayOutputStream;
+import java.nio.charset.StandardCharsets;
 import java.nio.charset.UnsupportedCharsetException;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/scm/SubversionSCM.java
```diff
@@ -3324,6 +3318,8 @@ public FormValidation doCheckLocal(@QueryParameter String value) throws IOExcept
      * Enables trace logging of Ganymed SSH library.
      * <p>
      * Intended to be invoked from Groovy console.
+     * @deprecated
```
olamy, line 3321: I  would even remove this (and so dependency on trilead-api) as the class in trilead-api is even already deprecated. 
https://github.com/jenkinsci/trilead-ssh2/blob/d9e5b647ca51e65db86e29a642f85eaee560bd10/src/com/trilead/ssh2/DebugLogger.java#L11

MarkEWaite:
I've confirmed that incremental release 2.17.4-rc1255.4e4a_35243fd5 (from this pull request) resolves the test failure in the plugin bill of materials.  Thanks @olamy !

MarkEWaite (approved):

jglick (approved):
@didiez are you still maintaining this? A release is needed. (I have write permissions it seems but would rather not use them. Also #277 would be nice.)

timja:
I'm not maintaining this functionally but I'll help with ecosystem related fixes if there's a working PR provided like in this case.

PR #341, opened 2025-02-12
Allow '.' in usernames

buminks:
Allow '.' in a username as valid character in exclude username filter.

The same kind of change was done in #233.

### Testing done

mvn verify is running without errors.

before the change:
image

after the change:
image (1)

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/hudson/scm/SubversionSCM.java
+++ b/src/main/java/hudson/scm/SubversionSCM.java
@@ -2502,9 +2502,9 @@ public class SubversionSCM extends SCM {
 
         /**
          * Regular expression for matching one username. Matches 'windows' names ('DOMAIN&#92;user') and
-         * 'normal' names ('user'). Where user (and DOMAIN) has one or more characters in 'a-zA-Z0-9_-')
+         * 'normal' names ('user'). Where user (and DOMAIN) has one or more characters in 'a-zA-Z0-9_.-')
          */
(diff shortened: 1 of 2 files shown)


Eccenux:
+1 from me.

It should work with basic auth which is implemented in (dot should work in any encoding):
`svnkit\src\main\java\org\tmatesoft\svn\core\internal\io\dav\http\HTTPBasicAuthentication.java`

Fine in NTLM too
https://learn.microsoft.com/en-us/previous-versions//cc722458(v=technet.10)?redirectedfrom=MSDN
> User names cannot contain certain characters. Invalid characters include:
> `" / \ [ ] : ; | = , + * ? < >`
> User names can contain all other special characters, including spaces, periods, dashes, and underscores. But it's generally not a good idea to use spaces in account names.

Eccenux (approved):
OK

buminks:
@Eccenux Thanks for your support.

PR #1597, opened 2026-09-02
Document that included/excluded region regexes match the full path #1431

AzazelSensei:
## Was this PR generated or assisted by AI?

NO

## Purpose of the pull request

Included/excluded region help said the fields use regular expressions, but not that each pattern has to match the whole Subversion path (`Matcher.matches()`). A relative path such as `src/main/.*` never triggers. I documented that and added a test for #1431 / JENKINS-45199.

## Brief change log

- Spell out full-path matching in the included and excluded region help
- Cover relative vs whole-path patterns in `DefaultSVNLogFilterTest`

## Verify this pull request

This change added tests and can be verified as follows:

- `DefaultSVNLogFilterTest.includedAndExcludedRegionsMatchEntirePath`

### DIFF
--- a/src/main/resources/hudson/scm/SubversionSCM/help-excludedRegions.html
+++ b/src/main/resources/hudson/scm/SubversionSCM/help-excludedRegions.html
@@ -4,6 +4,13 @@
   <p/>
   Each exclusion uses regular expression pattern matching, and must be separated by a new line.
   <p/>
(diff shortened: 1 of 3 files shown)


Eccenux (approved):
Looks OK, I like the examples. A bit of an odd pull request template, but it is fine.

PS: My friend was an Azazel in an RPG :)

